PR #5, opened 2023-12-12
Add generally applicable information for APIs

keirthana:


ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
-The documentation in this section provides a reference for the following APIs provided by Anbox Cloud:
+All communication between AMS and its clients happen using a RESTful API over HTTP. This API is encapsulated over either TLS (for remote operations) or a Unix socket (for local operations).
+
+## APIs and their structure
+Anbox Cloud provides the following APIs:
+
 * AMS HTTP API
 * Anbox Cloud HTTP API
```
adglkh, line 7: What's the `Anbox Cloud HTTP API`? do you mean the Anbox HTTP API that was referred below?

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
 * Stream Gateway API
 * Anbox Platform SDK API
 
-Also see [SDKs](https://discourse.ubuntu.com/t/anbox-cloud-sdks/17844) for a more detailed description of the SDKs that Anbox Cloud provides.
+All these APIs except for the [Anbox HTTP API](https://discourse.ubuntu.com/t/anbox-http-api/17819) have an auto-generated Swagger specification describing its API endpoints.
```
adglkh, line 11: The Anbox Platform SDK API is generated by in [doxygen](https://www.doxygen.nl/) actually

adglkh (changes requested):

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
+
+* `GET /1.0/artifacts`
+* `PATCH /1.0/instances/<name>`
+
+## API versioning
+The details of a version of the API can be retrieved using `GET /<version>`. For example, `GET /1.0`.
+
+If an API version is bumped to a major version, it indicates that backward compatibility is affected.
+
+Feature additions done without breaking backward compatibility only result in addition to `api_extensions` which can be used by the client to check if a given feature is supported by the server.
```
ajanon, line 31: I think we're missing an 's' here.
```suggestion
Feature additions done without breaking backward compatibility only result in additions to `api_extensions` which can be used by the client to check if a given feature is supported by the server.
```

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
+### Standard return value
+For a standard synchronous operation, the following dict is returned:
+
+```json
+{
+    "type": "sync",
+    "status": "Success",
+    "status_code": 200,
+    "metadata": {}               # Extra resource/action specific metadata
+}
```
ajanon, line 49: JSON does not support comments (i.e., parsers won't ignore what is after the '#')
I think it's reasonably clear what you want to say here with the comment, but it's technically incorrect. I'm not sure of a better way to do it, though.

It doesn't look really good right now I think:
image

Wdyt?

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
+```
+
+HTTP code must be 200.
```
ajanon, line 52: Why 'must'? It's not something the user has any control over, and it's not a specification to follow either.
How about 'is'?

Also, I think the correct term is 'HTTP response status code', see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status and https://httpwg.org/specs/rfc9110.html#overview.of.status.codes

```suggestion
HTTP response status code is 200.
```

Wdyt?

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
+
+### Background operation
+When a request results in a background operation, the HTTP code is set to 202 (Accepted) and the Location HTTP header is set to the operation URL.
+
+The body is a dict with the following structure:
```
ajanon, line 57:
```suggestion
The response body is a dict with the following structure:
```

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
+```json
+{
+    "type": "error",
+    "error": "Failure",
+    "error_code": 400,
+    "metadata": {}                      # More details about the error
+}
+```
+
+HTTP code must be one of 400, 401, 403, 404, 409, 412 or 500.
```
ajanon, line 104: See above.

ref/api-reference.md
```diff
@@ -1,7 +1,174 @@
...
+PUT replaces the entire object with a new definition, it's typically called after the current object state was retrieved through GET.
+
+To avoid race conditions, the ETag header should be read from the GET response and sent as If-Match for the PUT request. Doing so makes the request fail if the object was modified between GET and PUT.
+
+PATCH can be used to modify a single field inside an object by only specifying the property that you want to change. To unset a key, setting it to empty will usually do the trick, but there are cases where PATCH won't work and PUT needs to be used instead.
+
+## Authorisation
+Some operation may require a token to be included in the HTTP Authorisation header even if the request is already authenticated using a trusted certificate. If the token is not valid, the request is rejected by the server. This ensures that only authorised clients can access those endpoints.
+
+    `Authorization: bearer <token>`
```
ajanon, line 163:
```suggestion
    Authorization: bearer <token>
```

This should already be formatted with a monospaced font as it is indented to the right.

ajanon (changes requested):

ref/api-reference.md
keirthana: Yes, by terminology convention, it is better to avoid using just "Anbox" anywhere and use the full product name "Anbox Cloud". I have corrected both instances to read "Anbox Cloud HTTP API".

ref/api-reference.md
keirthana: I have made the sentence more generic.

ref/api-reference.md
keirthana: Agreed, I tested use "//" but that didn't work well either. For now, I am removing the comments since it seems fairly straight-forward.

PR #21, opened 2024-01-16
feat(ref): add VHAL HTTP API documentation

ajanon:
This adds VHAL endpoint documentation to the HTTP API docs based on the current version of the VHAL implementation.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -115,6 +115,7 @@ Graylog
 gRPC
 gzip
 HAProxy
+HIDL
 hotfix
 HTTPS
 HW
@@ -180,6 +181,7 @@ nvidia
 NVMe
 OAuth
 OBB
+OBD
 observability
 OES
 OOB
@@ -287,6 +289,7 @@ validator
 vCPU
 vCPUs
 VHAL
+VhalPropertyStatus
 viewport
 VirGL
 virglrenderer
--- a/reference/anbox-https-api.md
+++ b/reference/anbox-https-api.md
@@ -63,6 +63,8 @@ HTTP code must be one of 400 or 500.
    * [`/1.0/sensors`](#heading--10sensors)
    * [`/1.0/tracing`](#heading--10tracing)
    * [`/1.0/platform`](#heading--10platform)
+   * [`/1.0/vhal`](#heading--10vhal)
+      * [`/1.0/vhal/config`](#heading--10vhalconfig)
 
 ## API details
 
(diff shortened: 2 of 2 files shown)


morphis (approved):
LGTM

ref/anbox-https-api.md
```diff
@@ -491,3 +494,211 @@ Platform | Field name       | Available since   | JSON type | Access | Descripti
...
+To get the list of available properties and areas, query first the [`1.0/vhal/config` endpoint](#heading--10vhalconfig).
+
+Example return value for `curl -s -X GET --unix-socket /run/user/1000/anbox/sockets/api.unix s/1.0/vhal/0x15600503/0x31 | jq .`:
+
+```bash
+{
+  "metadata": {
+    "value": {
+      "area_id": 49,         # Requested area.
+      "bytes": [],           # Raw bytes value as an 8-bit unsigned integer array.
```
adglkh, line 525: Same question I left on the MP,  do we need to keep the field when its value is empty? Same for int32_values/int64_values and others.

ajanon: I will rework serialization in a follow-up MP.
For now, I prefer to have the documentation reflect what's actually in the code: if we omit the fields in the future, I will update the documentation and remove them (or explain that they may be omitted).

ref/anbox-https-api.md
```diff
@@ -491,3 +494,211 @@ Platform | Field name       | Available since   | JSON type | Access | Descripti
...
+
+```bash
+{
+    "status": "Success",
+    "status_code": 200,
+    "type": "sync"
+}
+```
+
+<a name="heading--10vhalconfig"></a>
```
keirthana, line 595: These HTML anchors are no longer necessary. The new layout has a page level TOC which is auto-generated. See https://anbox-cloud.io/docs/exp/production for example.

It's best to leave out these anchors as I have a clean up task for them anyway. It breaks headings and navigation in RTD.

ref/anbox-https-api.md
```diff
@@ -491,3 +494,211 @@ Platform | Field name       | Available since   | JSON type | Access | Descripti
...
+`MIXED` property types may have multiple of these values set at the same time, see
+[VHAL property types](https://source.android.com/docs/automotive/vhal/property-configuration#property-types).
+
+Status can be one of the following values, taken from the
+[VhalPropertyStatus](https://cs.android.com/android/platform/superproject/+/android10-release:hardware/interfaces/automotive/vehicle/2.0/types.hal;l=2700-2720)
+enumeration:
+
+|Name|Value|Description|
+|-|-|-|
+|Available|0|Property is available and behaving normally|
```
keirthana, line 555: Very small nitpick. Better to have punctuation at the end of all sentences or not have have them for all.

keirthana (changes requested):

ref/anbox-https-api.md
adglkh: That sounds good to me .

adglkh (approved):
LGTM

PR #22, opened 2024-01-17
Rename folders

keirthana:
Rename folders to have full name of the Diataxis categories
Update file paths and URLs as needed

This change will help us be more organized when we work on changes for moving to RTD.

### DIFF
--- a/index.md
+++ b/index.md
@@ -282,6 +282,9 @@ Thinking about using Anbox Cloud for your next project? [Get in touch!](https://
 | /docs/howto/container/view-log | /docs/howto/instance/logs |
 | /docs/usage/usecase-container-configuration | /docs/howto/instance/geographic-location |
 | /docs/howto/containers/backup-and-restore | /docs/howto/instance/backup-and-restore |
+| /docs/howto/stream/web-client | /docs/tut/stream-client |
+| /docs/howto/troubleshoot/collect-info | /docs/howto/troubleshoot/landing |
+| /docs/manage/managing-containers | /docs/exp/instances |
 | /docs/exp/gpu-support | /docs/explanation/gpus-instances |
 | /docs/howto/container/launch | /docs/howto/instance/create |
 | /docs/howto/stream/debug | /docs/howto/troubleshoot/streaming-issues |


morphis (approved):
LGTM

index.md
```diff
@@ -128,221 +128,172 @@ Thinking about using Anbox Cloud for your next project? [Get in touch!](https://
...
-| /docs/installation/validation | /docs/howto/install/validate |
-| /docs/installation/upgrading-from-previous-versions | /docs/howto/update/upgrade-anbox |
-| /docs/howto/manage/managing-images | /docs/howto/manage/images |
-| /docs/howto/manage/managing-addons | /docs/howto/addons/landing |
-| /docs/howto/manage/managing-logs | /docs/howto/manage/logs |
-| /docs/howto/manage/managing-ams-access | /docs/howto/manage/ams-access |
-| /docs/manage/managing-ams-access | /docs/howto/manage/ams-access |
-| /docs/installation/installation-application-registry | /docs/howto/aar/deploy |
-| /docs/usage/usecase-application-testing | /docs/howto/application/test |
-| /docs/usage/usecase-virtual-devices | /docs/howto/application/virtual-devices |
```
ajanon, line 295: Shouldn't we still keep these redirects, just in case somebody still has old bookmarks or references?

index.md
```diff
@@ -128,221 +128,172 @@ Thinking about using Anbox Cloud for your next project? [Get in touch!](https://
...
-| /docs/manage/capacity-planning | /docs/exp/clustering |
-| /docs/manage/gpu-support | /docs/exp/gpu-support |
-| /docs/exp/gpu-support | /docs/exp/gpus-instances |
-| /docs/manage/benchmarking-a-deployment | /docs/exp/benchmarking |
-| /docs/manage/streaming-android-applications | /docs/exp/application-streaming |
-| /docs/usage/usecase-port-android-application-to-anbox-cloud | /docs/exp/porting-issues |
-| /docs/installation/installation-requirements | /docs/requirements |
-| /docs/exp/porting-issues | /docs/howto/port/landing |
-| /docs/exp/benchmarking | /docs/howto/manage/benchmarks |
-| /docs/changelog | /docs/release-notes |
```
ajanon, line 329: Same for these ones.

index.md
```diff
@@ -128,221 +128,172 @@ Thinking about using Anbox Cloud for your next project? [Get in touch!](https://
...
 | /docs/howto/container/launch | /docs/howto/instance/create |
 | /docs/howto/stream/debug | /docs/howto/troubleshoot/streaming-issues |
 | /docs/howto/manage/web-dashboard | /docs/howto/dashboard/web-dashboard |
 | /docs/howto/manage/logs | /docs/howto/troubleshoot/logs |
-| /docs/ref/platforms | /docs/exp/platforms |
-| /docs/requirements | /docs/ref/requirements |
-| /docs/release-notes | /docs/ref/release-notes |
-| /docs/roadmap | /docs/ref/roadmap |
-| /docs/component-versions | /docs/ref/component-versions |
-| /docs/supported-versions | /docs/ref/supported-versions |
```
ajanon, line 339: And these ones.

ajanon (changes requested):

index.md
keirthana: Yes, I was initially thinking of keeping them but the redirects table keeps growing too long and these redirects are at least a year before (or closer to that range). So I thought it makes much more sense to keep the recent redirects. Maybe we can discuss this in the SU and see what others think as well?

keirthana: Yes, I was initially thinking of keeping them but the redirects table keeps growing too long and these redirects are at least a year before (or closer to that range). So I thought it makes much more sense to keep the recent redirects. Maybe we can discuss this in the SU and see what others think as well?

index.md
```diff
@@ -128,221 +128,175 @@ Thinking about using Anbox Cloud for your next project? [Get in touch!](https://
...
-| 2 | ref/hooks | [Hooks](https://discourse.ubuntu.com/t/hooks/28555)|
-| 2 | ref/webrtc-streamer | [WebRTC streamer](https://discourse.ubuntu.com/t/webrtc-streamer/30195)|
-| 2 | ref/prometheus | [Prometheus metrics](https://discourse.ubuntu.com/t/prometheus-metrics/19521)|
-| 2 | ref/perf-benchmarks | [Performance benchmarks](https://discourse.ubuntu.com/t/performance-benchmarks/24709)|
-| 2 | ref/license-information | [License information](https://discourse.ubuntu.com/t/license-information/36649) |
-| 2 | ref/glossary | [Glossary](https://discourse.ubuntu.com/t/glossary/26204)|
+| 1 | reference/landing | [Reference](https://discourse.ubuntu.com/t/reference/28828) |
+| 2 | reference/releases-versions | [Releases and versions](https://discourse.ubuntu.com/t/releases-and-versions/37993) |
+| 3 | reference/roadmap | [Release roadmap](https://discourse.ubuntu.com/t/release-roadmap/19359)|
+| 3 | reference/release-notes/release-notes | [Release notes](https://discourse.ubuntu.com/t/release-notes/17842)|
```
ajanon, line 134: `reference/release-notes/release-notes`

Is that intended? Shouldn't it be something like `landing`, as for the other sub directories?

keirthana: This is intended because technically the first "release-notes" is the folder name and the folder is mainly for convenience of grouping all the release notes together. "release-notes.md" is the file that actually is the release notes.

PR #33, opened 2024-02-01
feat(android): document how to use renderdoc with Android

morphis:

PR #34, opened 2024-02-02
Document Android system tracing

keirthana:
A first draft of the system tracing topic leveraged from the internal docs.

This will need additional changes to `index.md` and `howto/android/landing.md` that are part of this PR: https://github.com/canonical/anbox-cloud-docs/pull/33/files

### DIFF
--- a/howto/android/system-tracing.md
+++ b/howto/android/system-tracing.md
@@ -1,0 +1,48 @@
+This guide describes how to enable system tracing in Android using Anbox Cloud. 
+
+## Enable Android System Tracing
+
+1. To enable Android system tracing, add the `android.enable_tracing` flag into the feature file. In the Android container, run:
+
+   ```bash
+   echo "android.enable_tracing" >> /var/lib/anbox/features
+    ```
+ 2. Specify the category you want to trace in the Android system using the `debug.atrace.tags.enableflags` system property. The supported categories are defined in [trace.h](https://cs.android.com/android/platform/superproject/main/+/main:system/core/libcutils/include/cutils/trace.h;l=50?q=ATRACE_TAG_INPUT). 
+ 
+    For instance, to enable the graphics category, add the following to the `/var/lib/anbox/android.json` file:
+
+    ```json
+      {
+        "hardware": "ranchu",
+        "api_level": 31,
+        "properties": [ "debug.atrace.tags.enableflags=2" ]
+      }
+    ```
+ 3. Launch an Anbox session and initiate tracing via the HTTP endpoint. This will not only trace the Anbox session but also capture events aggregated under the category specified in the above system property.: 
+
+    ```bash
+    curl --unix-socket /run/user/1000/anbox/sockets/api.unix --request POST --data '{"enable":true}' s/1.0/tracing
+      ```   
+     After your desired tracing time, stop the tracing:
+    ```bash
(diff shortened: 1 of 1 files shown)


keirthana:
@adglkh I pulled out the content for system tracing with some modifications into this PR. Please leave your feedback.

howto/android/system-tracing.md
```diff
@@ -0,0 +1,48 @@
+This guide describes how to enable system tracing in Android using Anbox Cloud. 
+
+## Enable Android System Tracing
+
+1. To enable Android system tracing, add the `android.enable_tracing` flag into the feature file. In the Android container, run:
```
keirthana, line 5: > In the Android container

I changed the "development container" to "Android container". Is it correct? Or should this be "In the Anbox instance"?

howto/android/system-tracing.md
```diff
@@ -0,0 +1,48 @@
...
+ 
+    For instance, to enable the graphics category, add the following to the `/var/lib/anbox/android.json` file:
+
+    ```json
+      {
+        "hardware": "ranchu",
+        "api_level": 31,
+        "properties": [ "debug.atrace.tags.enableflags=2" ]
+      }
+    ```
```
ajanon, line 20: I think we need to reboot the instance after changing the android.json file, but we need to double-check this with @morphis or @adglkh.

howto/android/system-tracing.md
```diff
@@ -0,0 +1,48 @@
...
+ 3. Launch an Anbox session and initiate tracing via the HTTP endpoint. This will not only trace the Anbox session but also capture events aggregated under the category specified in the above system property.: 
+
+    ```bash
+    curl --unix-socket /run/user/1000/anbox/sockets/api.unix --request POST --data '{"enable":true}' s/1.0/tracing
+      ```   
+     After your desired tracing time, stop the tracing:
+    ```bash
+    curl --unix-socket /run/user/1000/anbox/sockets/api.unix --request POST --data '{"enable":false}' s/1.0/tracing
+      ```   
+      The resulting trace file is available in the response once tracing is disabled. The trace file should look similar to:
```
ajanon, line 30: So...  The file is not available then, the _path_to the file is available in the response.

howto/android/system-tracing.md
ajanon: Android runs in a container nested in an instance (container or VM), so that seems accurate to me.

ajanon (changes requested):

keirthana:
Putting this on hold as there is some dilemma about whether to include this for 1.21.

morphis:
> Putting this on hold as there is some dilemma about whether to include this for 1.21.

What I agreed with @adglkh was that we don't document this yet officially as it's use is rather harry and we haven't completed the implementation yet to be fully integrated to what people expect from Android.

keirthana:
@morphis I have updated the RN with the feedback so far. Could you have another look?

I think we need to add [AC-2071](https://warthogs.atlassian.net/browse/AC-2071), could you help me with the user impact for this one? We discussed this in SU before my vacation and I couldn't follow up with you on this.

keirthana:
Closing the PR without merging as this is not ready yet. (As discussed and confirmed with the team as on Feb 13, 2024)

PR #42, opened 2024-03-01
Update stream client tutorial

keirthana:
Based on the feedback we received. updated the
stream client tutorial to:

* Include information about downloading the Anbox streaming SDK
* Include information about using WebRTC
* Add related links


jat-canonical (approved):
LGTM

PR #48, opened 2024-03-14
Fixing AMS config missing deprecated column

keirthana:
(Need this PR to test the auto generation of AMS configuration tables)

The deprecated data about AMS config attributes are available in the source but it is not getting generated when being published. This PR is to fix it.


morphis (approved):
LGTM

PR #50, opened 2024-03-22
feat(instance): add name assignment

morphis:


howto/instance/create.md
```diff
@@ -71,6 +71,20 @@ Launch a raw instance by providing the image ID in the following command:
 
 See [Provided images](https://discourse.ubuntu.com/t/provided-images/24185) for a list of images that are available in Anbox Cloud.
 
+### Launch an instance with a specific name
+
+*since 1.22.0*
+
+To allow identifying and referencing instances easier you can give them a name at creation time. Names must be unique and can be used with all CLI commands.
```
ajanon, line 78: Could we add some more information on what characters can the name contain? Is it ASCII-only? UTF-8? Are some special characters disallowed? Is there a maximum length?
This may not fit the how-to section and may be better suited for the reference section, though.

ajanon (changes requested):

howto/instance/create.md
keirthana: Agree on the additional information required about the instance name. However, if there is just a little more information about the naming conventions, it should reside here along with the relevant topic for context. I doubt if naming convention for an instance has content enough to warrant a separate reference topic at the moment. 

"Names must be unique and can be used with all CLI commands." - This can also be included in the list of conditions that instance names must fulfill.

howto/instance/create.md
```diff
@@ -71,6 +71,20 @@ Launch a raw instance by providing the image ID in the following command:
...
+
+To create an instance with a name run
```
keirthana, line 80:
```suggestion
To create an instance with a name, run:
```

howto/instance/create.md
```diff
@@ -71,6 +71,20 @@ Launch a raw instance by providing the image ID in the following command:
...
+
+    amc launch --name=foo ...
+
+This will create an instance with the name "foo" which can then be used in other CLI commands, for example when opening a shell for interacting with the instance
```
keirthana, line 84:
```suggestion
This will create an instance with the name "foo" which can then be used in other CLI commands. For example, when opening a shell for interacting with the instance, the following command can be used:
```

keirthana (changes requested):

howto/instance/create.md
morphis: That is a good point but something we need to implement more generally IMHO as we have the same restriction applying for names of all objects. I can create a story to cover that.

keirthana (approved):
LGTM

ajanon (approved):
LGTM

PR #53, opened 2024-03-27
Document object status

keirthana:
Documenting status and descriptions for image, node, instance and application

### DIFF
--- a/explanation/applications.md
+++ b/explanation/applications.md
@@ -9,6 +9,20 @@ To run on the Anbox Cloud platform, applications must fulfil the following requi
 
 If your application fulfils these requirements but you are still having issues running it on Anbox Cloud, file a [bug report](https://bugs.launchpad.net/indore-extern/+filebug).
 
+## Possible application status
+
+The following table lists the different status that an application can have depending on its state and what each status means:
+
+| Application status | Description |
+|--------------------|-------------|
+| `initializing` | The application is being created. Wait until the application reaches the `ready` state. |
+| `ready` | The application is successfully created and ready to be used. |
+| `deleted` | The application is deleted and no longer available for use. |
+| `error` | The application has encountered an error. |
+| `unknown` | A possible error occurred and the real state of the application cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, use [`amc-showlog`](https://discourse.ubuntu.com/t/amc-command-reference-show-log/40792) to troubleshoot. If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
+
 ## Bootstrap process
 
 When [creating an application](https://discourse.ubuntu.com/t/create-an-application/24198) from a directory, a tarball, or a zip archive, AMS will perform a bootstrap process, which builds the application and synchronises it across all LXD nodes in the cluster. There are major benefits that the bootstrap process provides:
--- a/explanation/applications.md
+++ b/explanation/applications.md
@@ -21,7 +21,7 @@ The following table lists the different statuses that an application can have de
 | `error` | The application has encountered an error. |
 | `unknown` | A possible error occurred and the real state of the application cannot be determined. |
 
-If you encounter the `error` or the `unknown` status, see if you can identify the base instance and troubleshoot using the instance logs (See [How to view the instance logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329)). If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
+If you encounter the `error` or the `unknown` status, see if you can identify the base instance and troubleshoot using the instance logs (See [How to view the instance logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329)). If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant instance logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
 
 ## Bootstrap process
 
--- a/explanation/applications.md
+++ b/explanation/applications.md
@@ -9,19 +9,19 @@ To run on the Anbox Cloud platform, applications must fulfil the following requi
 
 If your application fulfils these requirements but you are still having issues running it on Anbox Cloud, file a [bug report](https://bugs.launchpad.net/indore-extern/+filebug).
 
-## Possible application status
+## Possible application statuses
 
-The following table lists the different status that an application can have depending on its state and what each status means:
+The following table lists the different statuses that an application can have depending on its state and what each status means:
 
 | Application status | Description |
 |--------------------|-------------|
 | `initializing` | The application is being created. Wait until the application reaches the `ready` state. |
 | `ready` | The application is successfully created and ready to be used. |
-| `deleted` | The application is deleted and no longer available for use. |
+| `deleted` | The application has been deleted and no longer available for use. |
 | `error` | The application has encountered an error. |
 | `unknown` | A possible error occurred and the real state of the application cannot be determined. |
 
-If you encounter the `error` or the `unknown` status, use [`amc-showlog`](https://discourse.ubuntu.com/t/amc-command-reference-show-log/40792) to troubleshoot. If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
+If you encounter the `error` or the `unknown` status, see if you can identify the base instance and troubleshoot using the instance logs (See [How to view the instance logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329)). If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
 
 ## Bootstrap process
 
--- a/explanation/images.md
+++ b/explanation/images.md
@@ -1,0 +1,17 @@
+An image represents the Anbox Cloud image available in the AMS node. An image is the base for an instance, which contains all necessary components like Anbox or the Android root file system. The image type can a container or a virtual machine.
+
+# Possible image status
+
+The following table lists the different status that an image can have depending on its state and what each status means:
+
+| Image status | Description |
+|--------------------|-------------|
+| `initializing` | The image is currently being created. |
+| `created` | The image was uploaded to AMS but not yet available on all LXD nodes. |
+| `available` | The image is present on the remote but not in the LXD cluster. |
+| `active` | The image is available on all LXD nodes. |
+| `deleted` | The image is deleted and no longer available for use. |
+| `error` | The image has encountered an error. |
+| `unknown` | A possible error occurred and the real state of the image cannot be determined. |
+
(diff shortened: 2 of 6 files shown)


explanation/applications.md
```diff
@@ -9,6 +9,20 @@ To run on the Anbox Cloud platform, applications must fulfil the following requi
...
 
+## Possible application status
+
+The following table lists the different status that an application can have depending on its state and what each status means:
+
+| Application status | Description |
+|--------------------|-------------|
+| `initializing` | The application is being created. Wait until the application reaches the `ready` state. |
+| `ready` | The application is successfully created and ready to be used. |
+| `deleted` | The application is deleted and no longer available for use. |
```
ajanon, line 20: Should we use the past tense instead? e.g.,
The application was deleted and is no longer available for use.

I'm not sure what's better, wdyt?

explanation/images.md
```diff
@@ -0,0 +1,17 @@
+An image represents the Anbox Cloud image available in the AMS node. An image is the base for an instance, which contains all necessary components like Anbox or the Android root file system. The image type can a container or a virtual machine.
```
ajanon, line 1:
```suggestion
An image represents the Anbox Cloud image available in the AMS node. An image is the base for an instance, which contains all necessary components like Anbox or the Android root file system. The image type can be a container or a virtual machine.
```

or maybe:

```suggestion
An image represents the Anbox Cloud image available in the AMS node. An image is the base for an instance, which contains all necessary components like Anbox or the Android root file system. The image type can be container or virtual machine.
```

explanation/instances.md
```diff
@@ -60,15 +60,18 @@ Beyond that, the instance will be removed from AMS either because you deleted it
...
-`started` | The instance is started and now booting. During the boot sequence, possible hooks are executed. Only when all hooks have been executed, the instance will switch to `running`.
-`running` | The instance is fully up and running.
-`stopped` | The instance is fully stopped and will be deleted by AMS.
-`deleted` | The instance is deleted and will be removed from the AMS database soon.
-`error` | An error occurred while processing the instance. The instance is stopped. Further information about the error can be viewed with `amc show <instance id>`.
+| Status            |  Description |
+|-------------------|--------------|
+| `created`         | AMS has created an internal database object for the instance and will next schedule the instance onto a suitable LXD node. |
+| `prepared`        | AMS has decided the LXD node on which it will schedule the instance. |
+| `started`         | The instance is started and now booting. During the boot sequence, possible hooks are executed. Only when all hooks have been executed, the instance will switch to `running`. |
```
ajanon, line 67:
```suggestion
| `started`         | The instance is started and now booting. During the boot sequence, possible hooks are executed. Once all hooks have been executed, the instance will switch to `running`. |
```

I think it reads a bit better, and I think it carries the same meaning. Wdyt?

explanation/images.md
```diff
@@ -0,0 +1,17 @@
...
+|--------------------|-------------|
+| `initializing` | The image is currently being created. |
+| `created` | The image was uploaded to AMS but not yet available on all LXD nodes. |
+| `available` | The image is present on the remote but not in the LXD cluster. |
+| `active` | The image is available on all LXD nodes. |
+| `deleted` | The image is deleted and no longer available for use. |
+| `error` | The image has encountered an error. |
+| `unknown` | A possible error occurred and the real state of the image cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, use [`amc-showlog`](https://discourse.ubuntu.com/t/amc-command-reference-show-log/40792) to troubleshoot. If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
```
ajanon, line 17: Does that apply to images too? AFAICT `amc showlog` would only work on instances.

explanation/nodes.md
```diff
@@ -0,0 +1,18 @@
...
+| `created` | The node is successfully created and ready to be used. |
+| `initializing` | The status of the node when its corresponding LXD instance is currently being configured. |
+| `initialized` | The status of the node after its corresponding LXD instance is initialised. |
+| `online` | The node is available and can be used by the AMS. |
+| `offline` | The node is not available and cannot be used by the AMS. |
+| `deleted` | The node is deleted and no longer available for use. |
+| `error` | The node cannot be used because of an error. |
+| `unknown` | A possible error occurred and the real state of the node cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, use [`amc-showlog`](https://discourse.ubuntu.com/t/amc-command-reference-show-log/40792) to troubleshoot. If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
```
ajanon, line 18: Same as for images: does that apply to nodes too? AFAICT `amc showlog` would only work on instances.

explanation/applications.md
```diff
@@ -9,6 +9,20 @@ To run on the Anbox Cloud platform, applications must fulfil the following requi
 
 If your application fulfils these requirements but you are still having issues running it on Anbox Cloud, file a [bug report](https://bugs.launchpad.net/indore-extern/+filebug).
 
+## Possible application status
```
ajanon, line 12: Should it be plural?

ajanon (changes requested):
Some questions and comments.

explanation/applications.md
ajanon: i.e., 'Possible application statuses'?

explanation/images.md
keirthana: My understanding was if we use `amc showlog`, the instance log might have something to show what's wrong with the other objects related to it as well? Or is that not the case?

explanation/applications.md
keirthana: That is a widely debated topic, some sources say "status", some say "statuses" and some say both are fine. I am trying to find a better source to rely on.

explanation/images.md
ajanon: I don't think so, but we would need to double-check with the others.

keirthana: Corrected as per MM conversation.

ajanon (approved):
LGTM

explanation/instances.md
```diff
@@ -56,19 +56,22 @@ When an instance is stopped, it executes the following steps in order:
...
+
+An instance moves through different stages and correspondingly can have the following statuses depending on its current state.
+
+| Status            |  Description |
+|-------------------|--------------|
+| `created`         | AMS has created an internal database object for the instance and will next schedule the instance onto a suitable LXD node. |
+| `prepared`        | AMS has decided the LXD node on which it will schedule the instance. |
+| `started`         | The instance is started and now booting. During the boot sequence, possible hooks are executed. Once all hooks have been executed, the instance will switch to `running`. |
+| `running`         | The instance is fully up and running. |
+| `stopped`         | The instance is fully stopped and will be deleted by AMS. |
```
jat-canonical, line 69: This is not necessarily true we can also do an `amc stop <cid>` to stop an instance and restart it again?

jat-canonical (changes requested):

explanation/applications.md
```diff
@@ -9,6 +9,20 @@ To run on the Anbox Cloud platform, applications must fulfil the following requi
...
+
+| Application status | Description |
+|--------------------|-------------|
+| `initializing` | The application is being created. Wait until the application reaches the `ready` state. |
+| `ready` | The application is successfully created and ready to be used. |
+| `deleted` | The application has been deleted and no longer available for use. |
+| `error` | The application has encountered an error. |
+| `unknown` | A possible error occurred and the real state of the application cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, see if you can identify the base instance and troubleshoot using the instance logs (See [How to view the instance logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329)). If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
```
adglkh, line 24: The above link points people to view the stored instance logs rather than AMS logs, so 
with the [relevant AMS logs]
->
with the [relevant instance logs]

(In our documentation, we seem not to mention how to dump the AMS logs as end users usually won't need to do so. If there is a need, that probably means something bad happens. In that case, we could ask bug reporter to dump ams log further. however for this section, to me, that part can be omitted.)

explanation/images.md
```diff
@@ -0,0 +1,17 @@
...
+|--------------------|-------------|
+| `initializing` | The image is currently being created. |
+| `created` | The image was uploaded to AMS but not yet available on all LXD nodes. |
+| `available` | The image is present on the remote but not in the LXD cluster. |
+| `active` | The image is available on all LXD nodes. |
+| `deleted` | The image has been deleted and no longer available for use. |
+| `error` | The image has encountered an error. |
+| `unknown` | A possible error occurred and the real state of the image cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
```
adglkh, line 17: Viewing the instance logs to determine why an image ended up to `error` or `unknown` status won't help. In such a case, a piece of ams log would be helpful.

explanation/instances.md
```diff
@@ -56,19 +56,22 @@ When an instance is stopped, it executes the following steps in order:
...
+| `started`         | The instance is started and now booting. During the boot sequence, possible hooks are executed. Once all hooks have been executed, the instance will switch to `running`. |
+| `running`         | The instance is fully up and running. |
+| `stopped`         | The instance is stopped. If you wish to restart the instance, use `amc stop <instance_id>`.|
+| `deleted`         | The instance has been deleted and will be removed from the AMS database soon. |
+| `error`           | An error occurred while processing the instance. The instance is stopped. |
+| `unknown`         | A possible error occurred and the real state of the instance cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, use [`amc show <instance_id>`](https://discourse.ubuntu.com/t/amc-command-reference-show/40793) or [`amc-showlog`](https://discourse.ubuntu.com/t/amc-command-reference-show-log/40792) to troubleshoot. If you are still unable to figure out the issue, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
```
adglkh, line 74: Same above

explanation/nodes.md
```diff
@@ -0,0 +1,18 @@
...
+| `created` | The node is successfully created and ready to be used. |
+| `initializing` | The status of the node when its corresponding LXD instance is currently being configured. |
+| `initialized` | The status of the node after its corresponding LXD instance is initialised. |
+| `online` | The node is available and can be used by the AMS. |
+| `offline` | The node is not available and cannot be used by the AMS. |
+| `deleted` | The node has been deleted and no longer available for use. |
+| `error` | The node cannot be used because of an error. |
+| `unknown` | A possible error occurred and the real state of the node cannot be determined. |
+
+If you encounter the `error` or the `unknown` status, [file a bug](https://bugs.launchpad.net/anbox-cloud) with the [relevant AMS logs](https://discourse.ubuntu.com/t/how-to-view-the-instance-logs/24329#view-stored-logs-2).
```
adglkh, line 18: Same here

adglkh (changes requested):

explanation/applications.md
keirthana: Simplified it to instance logs for applications and instances and just to file a bug for images and nodes.

explanation/instances.md
```diff
@@ -66,12 +66,12 @@ An instance moves through different stages and correspondingly can have the foll
 | `prepared`        | AMS has decided the LXD node on which it will schedule the instance. |
 | `started`         | The instance is started and now booting. During the boot sequence, possible hooks are executed. Once all hooks have been executed, the instance will switch to `running`. |
 | `running`         | The instance is fully up and running. |
-| `stopped`         | The instance is fully stopped and will be deleted by AMS. |
+| `stopped`         | The instance is stopped. If you wish to restart the instance, use `amc stop <instance_id>`.|
```
jat-canonical, line 69: There is still a typo in this, it should be `amc start` to restart the instance. Either way should we even include this here ? why not just say that the instance is stopped and can be restarted if required...

jat-canonical (changes requested):

explanation/instances.md
keirthana: Oops, sorry. It should be fixed now.

jat-canonical (approved):
LGTM

adglkh (approved):
LGTM

PR #55, opened 2024-04-02
Document COS integration

keirthana:
- Add howto guide for COS integration
- Update landing page
- Update index page to include navigation link

### DIFF
--- a/howto/monitor/cos.md
+++ b/howto/monitor/cos.md
@@ -1,0 +1,8 @@
+Anbox Cloud uses [Canonical Observability Stack(COS)](https://charmhub.io/topics/canonical-observability-stack) to enable users to collect and monitor their workloads.
+
+Since Anbox Cloud uses Juju for deployment, by supporting [COS Lite](https://charmhub.io/topics/canonical-observability-stack/editions/lite) integration, we can export observability information with Prometheus, Loki and Grafana.
+
+Currently, only the [AMS charm](https://charmhub.io/ams) supports COS Lite integration. To gather observability metrics for the AMS charm,
+
+1. Deploy the COS Lite bundle. See the [COS Lite tutorial](https://charmhub.io/topics/canonical-observability-stack/tutorials/install-microk8s) for instructions.
+2. Deploy the AMS charm as described in the [Instrumenting machine charms](https://charmhub.io/topics/canonical-observability-stack/tutorials/instrumenting-machine-charms) tutorial.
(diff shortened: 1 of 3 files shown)


howto/monitor/cos.md
```diff
@@ -0,0 +1,8 @@
+Anbox Cloud uses [Canonical Observability Stack(COS)](https://charmhub.io/topics/canonical-observability-stack) to enable users to collect and monitor their workloads.
+
+Since Anbox Cloud uses Juju for deployment, by supporting [COS Lite](https://charmhub.io/topics/canonical-observability-stack/editions/lite) integration, we can export observability information with Prometheus, Loki and Grafana.
+
+Currently, only the [AMS charm](https://charmhub.io/ams) supports COS Lite integration. To gather observability metrics for the AMS charm,
+
+1. Deploy the COS Lite bundle. See the [COS Lite tutorial](https://charmhub.io/topics/canonical-observability-stack/tutorials/install-microk8s) for instructions.
+2. Deploy the AMS charm as described in the [Instrumenting machine charms](https://charmhub.io/topics/canonical-observability-stack/tutorials/instrumenting-machine-charms) tutorial.
```
morphis, line 8: That tutorial is quite confusing as it mixes many things (extending a charm with observability, deploying it and integrating with the cos stack). I am not sure if that serves the need we have here well enough to show how the charm can be related to the COS stack. Wdyt?

howto/monitor/cos.md
```diff
@@ -0,0 +1,8 @@
+Anbox Cloud uses [Canonical Observability Stack(COS)](https://charmhub.io/topics/canonical-observability-stack) to enable users to collect and monitor their workloads.
+
+Since Anbox Cloud uses Juju for deployment, by supporting [COS Lite](https://charmhub.io/topics/canonical-observability-stack/editions/lite) integration, we can export observability information with Prometheus, Loki and Grafana.
+
+Currently, only the [AMS charm](https://charmhub.io/ams) supports COS Lite integration. To gather observability metrics for the AMS charm,
```
morphis, line 5: We have to be careful here as this is only true for an unreleased version of the charm which we do not officially support yet. We need to call this out here.

howto/monitor/cos.md
```diff
@@ -0,0 +1,8 @@
+Anbox Cloud uses [Canonical Observability Stack(COS)](https://charmhub.io/topics/canonical-observability-stack) to enable users to collect and monitor their workloads.
```
morphis, line 1: I think we need to differentiate two parts here:

1. General machine monitoring with grafana-agent
2. Individual workload monitoring

As we can not provide 2) for an official release of Anbox Cloud yet, maybe we focus on 1) only?

morphis (changes requested):

howto/monitor/cos.md
jat-canonical: I still think majority of that tutorial is applicable as the customer needs to know about the grafana agent charm to deploy it first and then relate to the anbox charms. The only thing that is not of their use it point 2 which actually mentions changes to charm code which they are not expected to do. But apart from that everything still stands. So we can copy the information from there make it specfic to anbox and remove the information asking them to make changes to charm ?

keirthana:
Closing without merging as per discussion in https://warthogs.atlassian.net/browse/AC-2103. We will reopen this when we can officially support the integration in latest/stable

PR #59, opened 2024-04-10
fix(exp,ref): swap container and vm minimum disk size

ajanon:
In the docs, the minimum disk size for containers and VMs was swapped: the minimum disk size for a container is 3GB, but the minimum disk size for a VM is 15GB.


adglkh (approved):
LGTM

keirthana (approved):
LGTM!

PR #62, opened 2024-04-23
Document public messaging for custom images

keirthana:
For https://warthogs.atlassian.net/browse/AC-2458


explanation/custom-images.md
```diff
@@ -0,0 +1,8 @@
+Building custom images of Anbox Cloud based on AAOS is possible but may get tricky.
+
+[note type="information" status="Note"]Building custom images is possible only for Android 13 and is currently only possible for AAOS images only.[/note]
+
+For example, using customised Anbox Cloud images will lead to more maintenance work, missing security updates provided by the Anbox Cloud team or requiring manual intervention to deploy Anbox Cloud. It is for these reasons that we generally don't recommend using a custom image unless you have a real business need for it.
+
+If you have a use case that demands an Anbox Cloud image with customisations applied, please write to us through the users forum.
```
ajanon, line 7: Should we keep it more generic? i.e.:
```suggestion
If you have a use case that demands an Anbox Cloud image with customisations applied, please contact us.
```

Possibly with a link to a contact form or something.

explanation/custom-images.md
```diff
@@ -0,0 +1,8 @@
+Building custom images of Anbox Cloud based on AAOS is possible but may get tricky.
```
ajanon, line 1: How about saying that it is not something we recommend?

explanation/custom-images.md
```diff
@@ -0,0 +1,8 @@
...
+
+[note type="information" status="Note"]Building custom images is possible only for Android 13 and is currently only possible for AAOS images only.[/note]
+
+For example, using customised Anbox Cloud images will lead to more maintenance work, missing security updates provided by the Anbox Cloud team or requiring manual intervention to deploy Anbox Cloud. It is for these reasons that we generally don't recommend using a custom image unless you have a real business need for it.
```
ajanon, line 5: I would suggest being more direct to the difficulties the users may face, e.g.,
```suggestion
Using customised Anbox Cloud images will lead to more maintenance work on your end:
- you will be missing security updates provided by the Anbox Cloud team
- your custom images will need to be updated regularly to stay compatible with Anbox cloud
- manual intervention will be required to deploy Anbox Cloud

It is for these reasons that we generally don't recommend using a custom image unless you have a real business need for it.
```

I think it's important that we make it clear that this has a lot of drawbacks and costs for a customer, and they shouldn't try unless they have the resources to keep up.

ajanon (changes requested):

explanation/custom-images.md
adglkh: Also, do we want to make it clear to the end user that it's a **system** image?

ajanon: We would need to explain what's a system image then, I think it's better not to go down that rabbit hole.

explanation/custom-images.md
keirthana: I have used the same contact form that is on our product home page.

ajanon (approved):
LGTM, thanks for the update!

PR #63, opened 2024-04-23
Remove topic for enabling VirGL

keirthana:
VirGL is going to be enabled by default since 1.22 and there will be no necessity to enable it
explicitly. Hence this commit removes the topic
that was added earlier to enable VirGL.


ajanon (commented):
Just a small question : is there someplace in the docs where we will say that VirGL is the default now, or will that be only in the release notes?

ajanon:
> > Just a small question : is there someplace in the docs where we will say that VirGL is the default now, or will that be only in the release notes?
> 
> A: That's my thinking as well. If there is a small section mentioning this, perhaps we could refine it to say, "To fallback to the old renderer emugl, please enable the feature 'emugl.enabled' upon application creation." Meanwhile, we may want to announce the deprecation of emugl since the 1.23 release if we all agree on this.

I think we shouldn't mention that feature flag as it's supposed to be internal only. Better keep it vague.

keirthana:
> Just a small question : is there someplace in the docs where we will say that VirGL is the default now, or will that be only in the release notes?

I am wondering if it makes sense to add it to the [rendering architecture topic ](https://anbox-cloud.io/docs/explanation/rendering-architecture) as a note. If we agree on this, is there any difference it the default enabling for NVIDIA vs Intel, AMD GPUs or is it the same for all irrespective of GPUs?

keirthana:
> > Just a small question : is there someplace in the docs where we will say that VirGL is the default now, or will that be only in the release notes?
> 
> I am wondering if it makes sense to add it to the [rendering architecture topic ](https://anbox-cloud.io/docs/explanation/rendering-architecture) as a note. If we agree on this, is there any difference it the default enabling for NVIDIA vs Intel, AMD GPUs or is it the same for all irrespective of GPUs?

Got this clarified - this applies only for NVIDIA GPUs.

ajanon (approved):
LGTM, thanks!

explanation/rendering-architecture.md
```diff
@@ -22,7 +22,9 @@ As the Venus Vulkan driver only provides Vulkan API support for Android, we use
 
 In terms of performance, this could be perceived to have some transmission overhead when compared to the rendering on Intel and AMD GPUs. However, Anbox Cloud is optimised to keep this overhead minimal and the additional overhead due to the transmission of Vulkan API calls from the Android space to the Ubuntu side renderer is not significant enough to affect most use cases.
 
-In older versions Anbox Cloud used a similar rendering pipeline for NVIDIA GPUs which only supports OpenGL ES and EGL but cannot provide Vulkan API support.
+In older versions, Anbox Cloud used a similar rendering pipeline for NVIDIA GPUs which only supports OpenGL ES and EGL but cannot provide Vulkan API support.
+
+Starting 1.22.0, Anbox Cloud uses VirGL as the default renderer for NVIDIA GPUs. If VirGL is not compatible for your use case (and you require the older implementation), [contact us](https://anbox-cloud.io/contact-us).
```
adglkh, line 28: The above section was not accurate to me. We have supported two renderers EmuGL and VirGL since 1.21 release however back then, Emugl was used by default while people had to explicitly enable with `virgl.enable` feature to use the VirGL renderer, which is not the case anymore since Anbox Cloud 1.22.0, where VirGL is used by default instead and EmuGL is out of support since 1.23 release.

adglkh (changes requested):

jat-canonical (approved):
LGTM

explanation/rendering-architecture.md
adglkh: Sorry, I didn't make it clear. 
In the above section(especially the 1st sentence), from the reader's perspective,  what I got was that Anbox Cloud has been using the EmuGL from the beginning and prior to the anbox cloud 1.22 release, only emugl pipeline is supported by Anbox Cloud for the NVIDIA GPUs while that's not true since the virgl is available to use and supported since 1.21 release(and it's just not the default renderer back then).

> Right now, we are changing the default renderer and we want the users to know that may break things right?

A: +1, because of this, in case something bad happens under the virgl mode, the feature `emugl.enable` exists for people to switch back to the old emugl renderer.

keirthana: I have added a statement that explains VirGL was available since 1.21.0 but not the default option. About the feature flag to enable emuGL, @morphis confirmed in one of the standups that we don't want to document that now but instead have people contact us if they have a use case that does not work with VirGL. So I am keeping the rest of the content as is.

keirthana: @adglkh Could you review this and confirm if the changes are okay now?

adglkh: Yes, thanks for this.

PR #65, opened 2024-04-24
Document object name restrictions

keirthana:
As part of this PR, I have also moved the features to the Anbox supported features topic as it seems like a much better place to host the content.


reference/anbox-features.md
```diff
@@ -9,3 +9,100 @@ The following table lists some Anbox features and whether they are supported for
 | [Custom Android ID](https://discourse.ubuntu.com/t/ams-configuration/20872#custom-android-id-10)                    | ✓    |      |
```
ajanon, line 9: This link should be updated to point below.

ajanon (changes requested):
One nitpick.
Similarly to my comment inline: should we update the links in the release notes which point to ams-configuration for feature flags? There is at least one such instance in the 1.21.0 release notes.

reference/anbox-features.md
keirthana: Done

keirthana:
> One nitpick. Similarly to my comment inline: should we update the links in the release notes which point to ams-configuration for feature flags? There is at least one such instance in the 1.21.0 release notes.

Thanks, I have updated the link to the new topic but the section targets aren't generated until published. So I guess directing to the topic should be good for now and later I will update it to the section.

ajanon (approved):
LGTM, thanks!

jat-canonical (approved):
LGTM

PR #69, opened 2024-05-08
Add updated streaming stack diagram and explanation

keirthana:
This change includes the new streaming stack diagram that illustrates both ways of application streaming. I have tried to work with what was existing to add the alternative implementation and also the explanation for it.

In the process, I have rearranged the topic to have better flow while explaining the different components of the streaming stack.


explanation/application-streaming.md
```diff
@@ -1,39 +1,57 @@
...
+### Stream gateway
 
-The following illustration shows how the Streaming Stack components interact with each other when creating a new streaming session:
+The stream gateway provides the initial place where a client and an instance can start a peer to peer discussion. These ‘chat rooms’ are called sessions and are central to the rest of the stack. The gateway manages these sessions in order to connect the user and the instance and while doing so, it selects and makes available the best machines for the user based on the availability and geographical location of the machines.
 
-![Streaming stack sequence|2400x1350](https://assets.ubuntu.com/v1/a14d6ead-application_streaming_stack.png)
+### Stream agent
 
-## Streaming an Application
+The stream agent is the middleware between the stream gateway and the Anbox Management Service (AMS), It is responsible for securing access to AMS, launching instances, relay status information and more.
```
ajanon, line 19:
```suggestion
The stream agent is the middleware between the stream gateway and the Anbox Management Service (AMS). It is responsible for securing access to AMS, launching instances, relaying status information and more.
```
Or:
```suggestion
The stream agent is the middleware between the stream gateway and the Anbox Management Service (AMS), it is responsible for securing access to AMS, launching instances, relaying status information and more.
```
Wdyt?

explanation/application-streaming.md
```diff
@@ -1,39 +1,57 @@
...
-The Stream Gateway exists to enable clients find each other through sessions.Sessions contain user data, application information and more, but most importantly, they provide an entry point for both the client and the instance to start the signalling process. Signalling is a process by which both peers exchange messages and establish optimal codecs, network routes and content types.
+A messaging queue to allow components to communicate.
+
+The following illustration shows how the streaming stack components interact with each other when streaming visual output from an instance:
+
+![Streaming stack sequence|2400x1350](https://assets.ubuntu.com/v1/f6b703b0-Cloud-streaming-stack_v4-1-1.png)
+
+## Streaming an application by creating a session
+
+While streaming an application, session creation can happen in two ways as illustrated in the diagram above. Until the 1.22.0 release, to create Anbox Cloud instances that can be streamed, a session needed to be created on the stream gateway first. Once the session is created, the instance creation will be triggered as detailed in the process below:
```
ajanon, line 39: The second sentence is a bit hard to read imo. I think it could be easier if it were the other way around, e.g.:
```suggestion
While streaming an application, session creation can happen in two ways as illustrated in the diagram above. Until the 1.22.0 release, a session needed to be created on the stream gateway first to be able to create Anbox Cloud instances that can be streamed. Once the session is created, the instance creation will be triggered as detailed in the process below:
```
I think it is a bit easier to read, but this is probably very subjective. Wdyt?
(also, I'm not sure if we need to add 'be able to')

explanation/application-streaming.md
```diff
@@ -43,6 +61,16 @@ When optimal ICE candidates have been selected and codec capabilities have been
 
 The streams can then be consumed to display the final content.
 
+## Streaming an application by creating a stream-enabled instance
+
+Starting from 1.22.0, Anbox Cloud offers a much simpler way to create instances that can be streamed. This is the alternative implementation depicted in the streaming stack diagram.
+
+You can request AMS to create an instance and AMS will be able to dynamically create a session for the newly launched instance by contacting the stream agent. The agent mediates communication between stream gateway and AMS and provides the necessary session information to the AMS. The AMS proceeds with successfully creating an instance that can be streamed.
+
+[note type="information" status="Note"]
```
ajanon, line 70: I wonder if that should be a warning instead, to make it more visible?

images/Cloud-streaming-stack_v4-1.pdf
ajanon, line 1: Why do we need that PDF file? AFAICT, we're only using the PNG file in the markdown.

images/Cloud-streaming-stack_v4-1-1.png
ajanon, line 1: Couple of comments on that file:

I think "prefered" should be spelled "preferred" in the first column, second box from the top

Why do we have some sentences written with the third person singular while others are not? e.g:
- 'Starts a session based...'  (first column, second box from the top)
- 'Start an instance...' (fifth column, first box from the top)

ajanon (changes requested):

explanation/application-streaming.md
morphis: I think keep it informative makes sense. It's not that something gets broken when streaming isn't enabled and it might be not on purpose. I see this more as a hint.

images/Cloud-streaming-stack_v4-1.pdf
morphis: @keirthana do we got the PNG from design as well or do you generated that yourself? If we got both from design we can drop the PDF. If we only got the PDF then we should keep both.

morphis (approved):
LGTM

images/Cloud-streaming-stack_v4-1-1.png
keirthana: I admit there are some inconsistencies but there are some scenarios where the component that is being considered does not "perform" the function but rather is sending an instruction to another component to perform the function. 

This is the case mostly in the stream attach implementation - for example, the user asks the AMS to create the instance via API calls, so create instance as a command is right there.

I will request the design team to correct the inconsistencies (they mainly arose because of we repurposed an existing diagram and multiple people have worked on it but thanks for catching that) where applicable.

images/Cloud-streaming-stack_v4-1.pdf
keirthana: I am told that we will get SVG, AI and PNG once we give sign off on the diagram. I will replace the PDF with the sources and PNG format alone.

explanation/application-streaming.md
keirthana: Keeping this as a note as is. Warning is better used for breaking stuff.

images/Cloud-streaming-stack_v4-1-1.png
ajanon: I wasn't entirely sure about these, thanks for the explanation!

Regardless of what we do with the verbs, I think it would be better to fix the spelling of preferred (unless I'm gravely mistaken, and it can also be spelled with a single 'r').

keirthana: I will get the preferred corrected, for a moment there, I was thinking whether it is a UK English vs US English thing but it is not.

keirthana: This is done now!

ajanon (approved):
LGTM

PR #70, opened 2024-05-09
docs(charms): add upgrade instructions to enable stream attach support for charms

jat-canonical:
This commit adds the instructions required to enable the new stream attach support to charms for customers upgrading from older versions of anbox-cloud bundle.


keirthana (changes requested):
Some minor suggestions

morphis (approved):
LGTM

We should link this in the release notes (cc @keirthana )

PR #74, opened 2024-05-17
Add steps to use Azure quickstart template

keirthana:
This PR is to add steps to deploy appliance using the Azure quick start template.

The manual steps are still retained for users who wish to deploy manually but it is added inside an expandable element to make the howto guide appear simple.


adglkh (approved):
LGTM

PR #76, opened 2024-05-24
fix(application): fix the resources presets when showing application details

adglkh:
Need to update the resource presets on display after fixing the missing `cpus`
field upon application creation


keirthana (approved):
LGTM, thanks!

morphis (approved):
LGTM

PR #77, opened 2024-05-24
Replace instance-type with resources in example manifests

keirthana:
Since `instance-type` is deprecated, it is a good idea to use `resources` in our examples to lead by example.
I have removed some obvious places where it is a straight-forward replacement but have not done the replacements in outputs to maintain technical accuracy (we still list instance type in outputs for backwards compatibility).

### DIFF
--- a/howto/application/test.md
+++ b/howto/application/test.md
@@ -93,7 +93,10 @@ In this example we use the following application `manifest.yaml`:
 
 ```
 name: app
-instance-type: a4.3
+resources:
+  cpus: 4
+  memory: 4GB
+  disk-size: 8GB
 ```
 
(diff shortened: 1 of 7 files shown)


howto/application/test.md
```diff
@@ -93,7 +93,10 @@ In this example we use the following application `manifest.yaml`:
 
 ```
 name: app
-instance-type: a4.3
+resources:
+  cpus: 4
+  memory: 4GB
+  disk-size: 8GB
```
ajanon, line 99: Is this intended? Shouldn't this be 3GB of memory and disk-size?

keirthana: Ah sorry, that was for a different example where more resources where required. Fixed it.

ajanon (approved):
LGTM

PR #78, opened 2024-05-27
Create AAOS tutorial

keirthana:


tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,51 @@
+This tutorial guides you through basic operations that you can do with an application based on an Android Automotive OS (AAOS) image. 
+
+In this tutorial, we will be specifically focusing on the following tasks:
+- Creating an application using an AAOS image
+- Stream the application you created with a specific screen resolution
+- Set the temperature of the automotive using the Vehicle Hardware Abstraction Layer (VHAL) controls
```
keirthana, line 6: Need your opinions on usage of automotive vs automobile:

Automotive indicates any type of vehicle (could include boats, planes etc) while automobile indicates a more specific four-wheeled vehicle that is used for transportation.

I have currently used automotive since the application type on the dashboard has that option but automobile seems like a better fit since automotive generally aligns more with the industry and automobile seems to align with the vehicle itself. What do others think?

tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,55 @@
...
+## Set automotive temperature using the VHAL panel
+
+Once the session is created, you can see the Android UI streaming. The controls on the right help you set specific automotive controls.
+
+Try setting the temperature by performing the following steps:
+
+- Select *All VHAL properties* on the right.
+- In the *Search* bar, enter *HVAC_TEMPERATURE_SET*. 
+- Click on the HVAC panel icon to open it.
+- Modify the temperatures for *Area 1* and *Area 2*
```
keirthana, line 49: @ajanon Do you think it would be useful if we include information about changing the units for Celsius vs Fahrenheit? I have linked the VHAL documentation later for a full list of properties but since there could be some confusion if the user is not used to a particular unit of temperature and wants to switch?

ajanon: That's a good question. Alternatively, for the future, we could also default our images to use Celsius instead of Fahrenheit. In the meantime, it may be a good idea to at least mention it, so that users are not confused.

jat-canonical (approved):
LGTM

tutorial/getting-started-aaos.md
keirthana: Decided in SU to keep it as automotive.

tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,64 @@
...
+Try setting the temperature by performing the following steps:
+
+- Select *All VHAL properties* on the right.
+- Click on the HVAC panel icon to open it.
+- In the *Search* bar, enter *HVAC_TEMPERATURE_DISPLAY_UNITS*. This indicates the unit of temperature.
+    - Update the value from 49 to 50. This updates the unit to Celsius.
+
+        [note type="information" status="Note"]
+        By default, the HVAC panel displays the temperature in Fahrenheit while the VHAL properties show values in Celsius. To avoid confusion, this step switches the HVAC panel display to Celsius.
+        [/note]
```
ajanon, line 53: Shouldn't this come before the step, though?

tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,64 @@
...
+- Modify the temperatures for *Area 1* and *Area 2*.
+- Click *Save*.
+
+As you modify the temperatures and save your changes, you can see the temperature updating on the HVAC panel.
+
+## Learn more
+
+Learn about the various [supported system properties in the VHAL](https://source.android.com/docs/automotive/vhal/system-properties).
+
+You can also the [Anbox HTTP API](https://discourse.ubuntu.com/t/anbox-http-api/17819#heading--10vhal) to adjust the properties using the CLI.
```
ajanon, line 64:
```suggestion
You can also use the [Anbox HTTP API](https://discourse.ubuntu.com/t/anbox-http-api/17819#heading--10vhal) to adjust the properties using the CLI.
```

ajanon (changes requested):
One typo and one question from me.

tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,64 @@
...
+## Stream the application with a specific screen resolution
+
+On the *Sessions* screen, click *Create session*.
+
+Select the application that you created earlier.
+
+Any of the *Screen resolution* options listed for the application should work but to set a custom resolution, select *Custom* from the drop-down and set the following values:
+```
+Width = 2304
+Height = 3072
```
morphis, line 30: This seem quite large, especially for software rendering/encoding this will cause high CPU load if no GPU is involved. What was the reason to choose this resolution?

morphis (changes requested):

tutorial/getting-started-aaos.md
ajanon: It's set to that amount because we cannot set the DPI. If we stick with lower resolutions, it becomes impossible to use the UI.

ajanon: For instance: 
image

Or: 
image



We could ask the UI team to change the default resolutions available, but the two that we have correspond to actual vehicle screens, so we thought they were good examples, hence why we just took one resolution and doubled it.

morphis: Hm, that is unfortunate. Once https://github.com/canonical/ams/pull/75 lands we should refine this and use a lower resolution. Fine to keep what we have for now.

morphis (approved):
LGTM

tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,78 @@
+This tutorial guides you through basic operations that you can do with an application based on an Android Automotive OS (AAOS) image. 
+
+In this tutorial, we will be focusing on the following tasks:
+- Creating an application using an AAOS image.
+- Stream the application you created with a specific screen resolution.
+- Set the temperature of the automotive using the Vehicle Hardware Abstraction Layer (VHAL) controls.
```
ajanon, line 6: Should we keep the same tense for all three points here? i.e. "Create" instead of "Creating"?

tutorial/getting-started-aaos.md
```diff
@@ -0,0 +1,78 @@
...
+
+It is important to understand that higher screen resolutions consume more resources.
+
+Select a desired *Frame rate*, for example, 60.
+
+Confirm with *Create session*.
+
+## Set automotive temperature using the VHAL panel
+
+The dashboard uses self-signed certificates. Before the stream starts, you might see a warning from your browser to accept the certificates manually. You will see the Android UI streaming once you accept the certificate and reload the page.
```
ajanon, line 57: Given that we ask users to have access to the appliance dashboard above, is this necessary?

ajanon (commented):
LGTM overall, I just have two small questions still.

tutorial/getting-started-aaos.md
keirthana: Thanks, that's a good point. I made it consistent but used it the other way around because it reads better in the context.

tutorial/getting-started-aaos.md
keirthana: I think yes, because in my experience, I was able to access the dashboard and was thinking everything is working but streaming failed because I needed to accept the certificate for stream gateway. 

Would it be better if I called it the "certificate for stream gateway"?

ajanon: I think it's fine in this case. I think it will also be easier with the new microbox snap: if I understand correctly, we will only need to accept a single certificate, and it won't be necessary to accept another certificate. In the meantime, let's keep it as-is.

ajanon (approved):
LGTM, thanks!

PR #79, opened 2024-05-28
feat(howto): add snap refresh to the prerequisites

ajanon:
Users need to have the latest appliance version before they can upgrade. To that end, when they update their packages, they should also take this time to upgrade their snaps, to ensure the update goes smoothly.


keirthana (approved):
LGTM!

PR #80, opened 2024-05-29
feat: update Android feature list

morphis:
Vulkan 1.3 and 64 bit only support are available since 1.22 so we can mark both as supported.


ajanon (approved):
LGTM

keirthana (approved):
LGTM

PR #81, opened 2024-05-31
fix(reference): remove av1 experimental anbox feature from list

ajanon:
This feature flag has been removed nearly a year ago in favor of automated codec selection, and as such should not be documented here anymore.


adglkh (approved):
LGTM

PR #82, opened 2024-06-03
fix: update supported versions

morphis:
It was missed after the 1.21 release to update the supported versions in the reference documentation.


keirthana (approved):
LGTM, thanks!

PR #84, opened 2024-06-05
Add CLA check

keirthana:
As we declare docs open for contributions through GH (we have started with open docs academy but this will be the case when we move to RTD), we need the CLA check in place so that we can be sure that the contributor has signed the CLA.


ajanon (approved):
LGTM

morphis (approved):
LGTM

PR #93, opened 2024-07-15
RTD: Add format for instance.network_proxy

ajanon:
This adds the format for the value of instance.network_proxy.

Fixes https://bugs.launchpad.net/anbox-cloud/+bug/2072510


ajanon:
The link health check is failing, but I'm not sure what I should do about it.

keirthana:
> The link health check is failing, but I'm not sure what I should do about it.

If it is not failing due to any links that you added (which is the case with this PR), please ignore it. I need to watch the checker for a few weeks to see if it is really worth having it enabled despite the random breaks.

keirthana (approved):
LGTM

PR #95, opened 2024-07-19
Remove outdated discourse links checker pre-commit hook

keirthana:
We no longer need URLs to point to discourse posts.


adglkh (approved):
LGTM

PR #102, opened 2024-07-26
Update glossary to be able to reuse terms

keirthana:
The changes in this PR enables us to use a term defined in the glossary anywhere and it will be automatically linked to the glossary definition.


ajanon (approved):
LGTM

PR #103, opened 2024-07-29
Configure link check to ignore anchors

keirthana:
The link check is flaky when it comes to checking anchors. It often reports working anchors as broken and is unpredictable.

This PR adds configuration to disable checking the anchors.


morphis (approved):
LGTM

PR #106, opened 2024-07-30
Update dashboard topics for 1.23.0

keirthana:
- Remove screenshots as they are not easy to maintain
- Use intuitive text with icons where necessary to indicate actions (using icons is needed as sometimes the icons are not intuitive by themselves)
- Update dashboard content to reflect the merge of sessions and instances views.


explanation/web-dashboard.md
```diff
@@ -10,8 +10,7 @@ The dashboard comes pre-installed when you deploy the full version of Anbox Clou
 The dashboard allows multiple views and management of various objects such as applications, instances and nodes to some extent. The operations available from the dashboard may be limited as compared to the CLI, however if you are new to the CLI, the dashboard can be a friendlier start to Anbox Cloud.
 
 * Applications view - Allows creation, streaming and management of applications.
-* Sessions view - Allows a list view of the sessions and their status.
-* Instances view - Allows a detailed view of instances and the applications or images that they are created from, along with management options for the instances.
+* Instances view - Allows a detailed view of instances and the applications or images that they are created from, along with management and streaming options for the instances.
```
jat-canonical, line 13: Would an `other` instead of just `streaming` suffice ? since we will also have the adb connect option there now.

howto/instance/create-instance.md
```diff
@@ -4,13 +4,13 @@ To launch an application or an image, Anbox Cloud creates an instance for it. To
 
 ## Using the dashboard
 
-Select *Create Instance* on the *Instances* view to launch an instance. The information required for launching an instance depends on whether you are creating the instance from an application or an image.
+*Create Instance* in the instances view to launch an instance. The information required for launching an instance depends on whether you are creating the instance from an application or an image.
```
jat-canonical, line 7: Should this be `Click *Create Instance*` ?

jat-canonical (approved):
A couple of nits. Otherwise LGTM

howto/application/create-application.md
```diff
@@ -48,7 +48,7 @@ resources:
 
 ## Using the dashboard
 
-To create an application using the dashboard, use the *Add application* button on the *Applications* view, enter the required and any optional details that you want to provide and select *Add application*. This screen also provides the option to customise your application manifest.
+To create an application using the dashboard, click *Create application* from the applications view, enter the required and any optional details that you want to provide and confirm with *Create application* again. This screen also provides the option to customise your application manifest.
```
acarege, line 51: I would suggest calling all pages what they are instead of a "view". From a design & architecture perspective everything that exists in the main navigation is a page: Applications, Instances, Nodes, Registry and Configuration. In the dashboard, whenever we refer to a page we explicitly call it a page. So having unity in the documentation about that would be less confusing to the user imo 👐

howto/application/stream-application.md
```diff
@@ -24,7 +24,9 @@ For example, to create an instance with a 1080p resolution, a frame rate of 60 a
 
 The dashboard has in-browser streaming capabilities through WebRTC. It uses the {ref}`sec-streaming-sdk`.
 
-You can start a streaming session for any of the successfully created applications. An application is ready to be streamed when its *Status* is *Ready*. When you select the *Start a new session* icon for an application, you can specify the desired streaming attributes such as the screen resolution, frame rate, screen orientation before launching the session.
+When creating an instance, make sure you select the *Enable Streaming* capability to be able to stream your application. You can also set the desired streaming attributes using the *Virtual display* options available for an instance.
+
+You can start a streaming session for any of the successfully created applications. Once the associated instance is created and ready, click *Join session* ( ![join session icon](/images/join-session-icon.png) ) to start the stream.
```
acarege, line 29: I'm confused, is this about the CLI or the dashboard? If so, the sessions or anything related to it doesn't exist in the dashboard anymore 🤔

howto/instance/create-instance.md
```diff
@@ -4,13 +4,13 @@ To launch an application or an image, Anbox Cloud creates an instance for it. To
...
 
 When you create the instance from an application, the attributes you define for the application decide most of the properties of the instance.
 
 When you create the instance from an image, you can define the properties of the instance during its creation.
 
-Once you *Create and Start* an instance by providing the necessary attributes, you can view the instance and its status on the *Instances* view.
+Once you *Create and Start* an instance by providing the necessary attributes, you can view the instance and its status in the instances list view.
```
acarege, line 13: Isn't this also the case when they only "create"?

explanation/web-dashboard.md
keirthana: Thanks, I still think it is important to give an idea about the kind of options but have made it a bit more non-specific.

howto/instance/create-instance.md
keirthana: I was trying to combine the button name with the action performed but I guess it does not read well. I have reworded to focus on the action alone.

howto/application/stream-application.md
keirthana: That's a good catch, I was checking the demo server when I wrote this but I checked it now and the icon tooltip says "Stream" indeed. Maybe I was looking at the wrong machine.

howto/instance/create-instance.md
keirthana: Yes, I have generalized the term to mean the action instead of the button.

ajanon (approved):
LGTM

howto/application/create-application.md
acarege: Sounds great to me, thank you Keirthana 🎊

PR #114, opened 2024-08-15
Add initial explanation for crypto used by AMS

morphis:


morphis:
@keirthana I've added this as a draft here for now as we need to discuss where to put this best. This form satisfies SEC0030 and will need to be repeated for all other services / components.

keirthana:
> @keirthana I've added this as a draft here for now as we need to discuss where to put this best. This form satisfies SEC0030 and will need to be repeated for all other services / components.

@morphis Yes, there should be a separate section for cryptography. Let me set that up through another PR. Before that, this is targeted for 1.24 right? In that case, I will set up a branch for 1.24 on the repo - I know we still have versioning to be spec'ed out but we might need something in place for current work.

morphis:
> > @keirthana I've added this as a draft here for now as we need to discuss where to put this best. This form satisfies SEC0030 and will need to be repeated for all other services / components.
> 
> @morphis Yes, there should be a separate section for cryptography. Let me set that up through another PR. Before that, this is targeted for 1.24 right? In that case, I will set up a branch for 1.24 on the repo - I know we still have versioning to be spec'ed out but we might need something in place for current work.

This is fine to go in any time, not tied to a release

explanation/security/crypto_ams.md
```diff
@@ -0,0 +1,31 @@
+(exp-security-crypto-ams)=
+# AMS
+
+AMS is using cryptographic technology for
```
keirthana, line 4:
```suggestion
Anbox Management Service (AMS) is using cryptographic technology for:
```

explanation/security/crypto_ams.md
```diff
@@ -0,0 +1,31 @@
+(exp-security-crypto-ams)=
+# AMS
```
keirthana, line 2:
```suggestion
# Cryptographic information: AMS 
```

explanation/security/crypto_ams.md
```diff
@@ -0,0 +1,31 @@
...
+* Mutual TLS based authentication
+* Token based authentication
+
+## TLS transport encryption
+
+All network endpoints exposed by the AMS server are secured with TLS using an 4096 bit RSA key. AMS strictly enforces TLS 1.3 or better and does not provide backward compatibility with older TLS versions.
+
+## Mutual TLS based authentication
+
+AMS authenticates with all of its clients through the use of mutual TLS. The server authenticates a client by the presented client certificate and checks if it is recorded in its trust store. A client checks the certificate provided by the server against a recorded one. Once both sides checked and accepted the presented certificate a TLS connection is established for API access.
```
keirthana, line 16:
```suggestion
AMS authenticates with all of its clients through the use of mutual TLS. The server authenticates a client by the presented client certificate and checks if it is recorded in its trust store. A client checks the certificate provided by the server against a recorded one. Once both sides have checked and accepted the presented certificate a TLS connection is established for API access.
```

explanation/security/crypto_ams.md
```diff
@@ -0,0 +1,31 @@
...
+
+AMS uses for its server an RSA key with a length of 4096 bits. The standard AMS client `amc` uses a RSA key with a length of 4096 bits. The key in both cases is generated by the use of the [`rsa.GenerateKey`](https://pkg.go.dev/crypto/rsa#GenerateKey) method as provided by the Go standard library.
```
keirthana, line 18:
```suggestion
AMS uses an RSA key with a length of 4096 bits for its server. The standard AMS client `amc` uses a RSA key with a length of 4096 bits. The key in both cases is generated by the use of the [`rsa.GenerateKey`](https://pkg.go.dev/crypto/rsa#GenerateKey) method as provided by the Go standard library.
```

explanation/security/crypto_ams.md
```diff
@@ -0,0 +1,31 @@
...
+
+Authenticated client scan add additional certificates to the trust store AMS manages. The trust stored keeps a copy of all registered certificates and stores them in an [etcd database](https://etcd.io/).
```
keirthana, line 20:
```suggestion
Authenticated clients can add additional certificates to the trust store AMS manages. The trust stored keeps a copy of all registered certificates and stores them in an [etcd database](https://etcd.io/).
```

explanation/security/crypto_ams.md
```diff
@@ -0,0 +1,31 @@
...
+
+Communication with LXD is using mutual TLS based authenication too. More details on this can be found in the [LXD documentation](https://documentation.ubuntu.com/lxd/en/latest/authentication/).
+
+## Token based authentication
+
+Individual Anbox instances have access to a limited set of API endpoints exposed by the AMS server to submit status information at runtime. Access is authenticated by a scope limited JWT based token. The token uses an expiry of one year and is signed with a HMAC using SHA-256 (HS256) and a 64 byte secret key. The [`jwt.New`](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#Token) method is being used internally to generate the JWT token.
```
keirthana, line 26:
```suggestion
Individual Anbox instances have access to a limited set of API endpoints exposed by the AMS server to submit status information during runtime. Access is authenticated by a scope-limited JWT based token. The token is valid for one year and is signed with a [HMAC](https://www.okta.com/identity-101/hmac/) using SHA-256 (HS256) and a 64 byte secret key. The [`jwt.New`](https://pkg.go.dev/github.com/golang-jwt/jwt/v5#Token) method is used internally to generate the JWT token.
```

keirthana (changes requested):
Some minor changes, otherwise looks good!

keirthana:
> > > @keirthana I've added this as a draft here for now as we need to discuss where to put this best. This form satisfies SEC0030 and will need to be repeated for all other services / components.
> > 
> > 
> > @morphis Yes, there should be a separate section for cryptography. Let me set that up through another PR. Before that, this is targeted for 1.24 right? In that case, I will set up a branch for 1.24 on the repo - I know we still have versioning to be spec'ed out but we might need something in place for current work.
> 
> This is fine to go in any time, not tied to a release

@morphis I had a closer look at this and I see that you have created the structure in place with the security folder. So let's go ahead with this and move things around when more information is available. Left a few suggestions to improve readability, otherwise LGTM.

keirthana (approved):
LGTM

ajanon (approved):
LGTM

PR #122, opened 2024-08-27
feat: add crpyto explanation for the dashboard

morphis:


explanation/cryptography/crypto_dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+4. The `decode()` function verifies the token's signature using the secret key and algorithm. If verification is successful, it decodes the token and returns the original payload dictionary.
+
+## User authentication
+
+When authenticating with an external identity provider the following actions are performed:
+
+* Retrieving ID Tokens: ID tokens, which contain the user's email and full name, are obtained from the authentication server.
+* Decoding Unverified Headers: Initially, the `get_unverified_header()` function from [python-jose](https://github.com/mpdavis/python-jose/) is used to decode the token's unverified headers and extract the "kid" (key ID).
+* Fetching and Caching JWKS: The JSON Web Key Set (JWKS) is fetched from the JWKS endpoint and cached. For subsequent requests, the cached JWKS is checked for a key matching the "kid" in the token. If the key is not found in the cache, the JWKS is refetched.
+* Decoding the Token: The `decode()` function from [python-jose](https://github.com/mpdavis/python-jose/) is used to decode the token. This process uses the JWKS containing the key ID and verifies the audience.
```
keirthana, line 39:
```suggestion
When authenticating with an external identity provider, the following actions are performed:

Retrieving ID Tokens
  : ID tokens, which contain the user's email and full name, are obtained from the authentication server.

Decoding Unverified Headers
  : Initially, the `get_unverified_header()` function from [python-jose](https://github.com/mpdavis/python-jose/) is used to decode the token's unverified headers and extract the key ID.

Fetching and Caching JWKS
  : The JSON Web Key Set (JWKS) is fetched from the JWKS endpoint and is cached. For subsequent requests, the cached JWKS is checked for a key matching the key ID in the token. If the key is not found in the cache, the JWKS is refetched.

Decoding the Token
  : The `decode()` function from [python-jose](https://github.com/mpdavis/python-jose/) is used to decode the token. This process uses the JWKS containing the key ID and verifies the audience.
```

keirthana: The list format reads better for this topic as there are multiple lists on this topic

explanation/cryptography/crypto_dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Mutual TLS based authentication
+
+The dashboard uses mutual TLS authentication to establish a trusted TLS communication channel with AMS to call its API. For this it generates a TLS certificate using an 4096 bit RSA key.
```
keirthana, line 43:
```suggestion
The dashboard uses mutual TLS authentication to establish a trusted TLS communication channel with the Anbox management Service (AMS) to communicate with the AMS API. To do this, the dashboard generates a TLS certificate using a 4096 bit RSA key.
```

explanation/cryptography/crypto_dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Registration of new users
+
+For registering a new user, a token is generated using the PyJWT library.
+
+* The token includes the following payload data: the user's email and the token's validity period.
+* The `payload` dictionary is passed to the `encode()` function from the PyJWT library.
+* A secret key is used to securely sign the token.
+* The [HMAC](https://www.okta.com/identity-101/hmac/) SHA-256 algorithm is used to sign the token.
+* The `encode()` function returns a JWT, which is a string representing the encoded and signed payload data.
```
keirthana, line 23: To follow the same format for both encoding and decoding
```suggestion
1. The token includes the user's email and the token's validity period as payload data. This `payload` dictionary is passed to the `encode()` function from the PyJWT library.
1. A secret key is used to sign the token. The [HMAC](https://www.okta.com/identity-101/hmac/) SHA-256 algorithm is used in the signing process.
1. The `encode()` function returns a [JWT](https://jwt.io/), which is a string representing the encoded and signed payload data.
```

explanation/cryptography/crypto_dashboard.md
```diff
@@ -0,0 +1,49 @@
...
+
+The dashboard uses [PyJWT](https://github.com/jpadilla/pyjwt) to decode the token after it has been presented by the user, verifies its validity and checks its expiration:
+
+1. The token (JWT string) is passed to the `decode()` function from the [PyJWT](https://github.com/jpadilla/pyjwt) library.
+2. The secret key is used to verify the token's signature.
+3. The expected signing algorithm for the token is [HMAC](https://www.okta.com/identity-101/hmac/) SHA-256.
+4. The `decode()` function verifies the token's signature using the secret key and algorithm. If verification is successful, it decodes the token and returns the original payload dictionary.
```
keirthana, line 30: Following same format as for encode:
```suggestion
1. The token, in the form of a JWT string, is passed to the `decode()` function from the PyJWT library.
1. The secret key is used to verify the token's signature and the expected signing algorithm for the token is [HMAC](https://www.okta.com/identity-101/hmac/) SHA-256.
1. The `decode()` function verifies the token's signature using the secret key and algorithm. If verification is successful, it decodes the token and returns the original payload dictionary.
```

keirthana (changes requested):

keirthana (approved):
LGTM, thanks!

PR #125, opened 2024-08-28
fix(howto): Change VNDK version for custom VHAL to 34

ajanon:
Since 1.23.0, we have switch to a new VNDK version, and we seemed to have missed updating related documentation.


ajanon:
@keirthana Do you think we should do something more here? Adding this to the release notes for 1.23.1 as something we missed maybe? Wdyt?

morphis (approved):
LGTM!

keirthana (approved):

PR #134, opened 2024-09-04
fix(ref): use correct image version for 1.23.0

morphis:
The versioning scheme changed starting with 1.23.0 and needs to be correctly reflected in the component versions.


adglkh (approved):
LGTM!

keirthana (approved):

PR #136, opened 2024-09-09
feat: add script to parse swagger generate from ams

jat-canonical:
This commit adds the script to parse the swagger generated via ams to render the node and ms configuration fields.


scripts/parse.py
```diff
@@ -0,0 +1,78 @@
...
+                        if base == "gpus":
+                            new_base = f"{base}.<id>"
+                        node_props[f"{new_base}.{subprop_name}"] = subprop_attrs
+            node_props.pop(base)
+        else:
+            val = node_props.pop(base)
+            node_props[base.replace("_", "-")] = val
+        if "description" in prop:
+            prop["description"] = prop["description"].replace("\n", " ")
+        return node_props
```
adglkh, line 74: hmm, is this right?
To me the indentation is wrong, which means we only handle the first prop, then returns the node_props?

scripts/parse.py
```diff
@@ -0,0 +1,78 @@
...
+    env = Environment(loader=FileSystemLoader("."))
+    templ = env.get_template("template.md.j2")
+    text = templ.render(configs=configs, nodes=nodes)
+    with open(args.output_file, mode="w+") as op:
+        op.write(text)
+
+
+def _parse_config_schema(swagger) -> dict:
+    ams_config_path = "/1.0/config"
+    config_props = swagger["paths"][ams_config_path]["get"]["responses"]["200"][
```
adglkh, line 40: This looks a bit weird.

adglkh (changes requested):

scripts/parse.py
jat-canonical: oh yes good catch thanks. I had refactore the code a bit at the end.

scripts/parse.py
jat-canonical: in what way ?

keirthana:
So does this take the input from the swagger.json available at https://canonical.github.io/anbox-cloud.github.com/latest/ams/?

scripts/parse.py
adglkh: The following looks decent to me 
```
config_pros = swagger["paths"][ams_config_path]["get"]["responses"]["200"]
        ["schema"]["properties"]["metadata"]["properties"]["config"]["properties"]
```
rather than the way it looked like above

jat-canonical: That is just `black` formatting this with I think the character limit restrictions.

adglkh: TBH, that looks odd.

adglkh (approved):
LGTM

conf.py
```diff
@@ -133,3 +133,5 @@
 if 'github_issues' in html_context and html_context['github_issues'] and not disable_feedback_button:
     html_js_files.append('github_issue_links.js')
 html_js_files.extend(custom_html_js_files)
+
+generate_ams_configuration()
```
keirthana, line 137: Minor nitpick: Can we add comments explaining this is an Anbox specific thing and that we need to add this line to conf.py everytime we upgrade to a new version of starter pack?

Ideally, we are never supposed to edit conf.py and if edits are needed, we are supposed to push it upstream. But this is such a project specific thing that I don't see value in pushing this to upstream. 
```suggestion
# Anbox specific function to generate dynamic AMS configuration
# Add this change to conf.py every time the starter pack is upgraded to a later version.
generate_ams_configuration()
```

keirthana (approved):
LGTM

PR #138, opened 2024-09-10
fix(ref): use correct version for 1.23.0 snaps

morphis:
The versioning scheme changed with 1.23.0 and the version is no longer just 1.23.0 but includes a suffix for the git revision.


keirthana (approved):
LGTM!

PR #141, opened 2024-09-10
feat(ref): add missing internal network ports for appliance

morphis:
We should document these as they are relevant for the user.


keirthana (approved):
LGTM

PR #147, opened 2024-09-18
Automate loading of AMS and Gateway API spec from swagger

keirthana:


morphis:
@keirthana let me know when this is ready. It looks really nice already but two things I noticed:

1. There are now two entries in the navigation saying "Main API specification"
2. In the dark theme the API spec doesn't look great

image

keirthana:
> @keirthana let me know when this is ready. It looks really nice already but two things I noticed:

It is ready for review. I have based it on the content automation branch to see how it responds on RTD and it looks good.
> 
>     1. There are now two entries in the navigation saying "Main API specification"
> 

I have updated the titles now.

>     2. In the dark theme the API spec doesn't look great

Agreed, but I am not sure if there is anything we can do about it for now, let me check some options.
> 
> 
> image

PR #150, opened 2024-09-23
feat(scripts): add field deprecation and doc hyperlink parsing in the script

jat-canonical:
This commit adds the feature to insert deprecated fields and other hyperlinks from the swagger.


keirthana:
Overall, looks good to me, although I'd prefer someone with a working knowledge of Python to review as well.

One question, this would require the [AMS PR](https://github.com/canonical/ams/pull/528) to go through first right?

keirthana (approved):

PR #152, opened 2024-09-25
feat(gh): add workflow to automatically update API specs

morphis:
This consume the anbox-cloud-appliance snap from <base version>/stable and imports the swagger API spec from AMS. If changes are detected, the workflow creates a PR for inclusion in the repository.

The workflow uses the anbox-cloud-appliance GitHub action to install and setup the appliance so we can retrieve the swagger API spec from the AMS API endpoint. For the gateway API the swagger is not yet served from the API so we cannot do the same here, yet.

As we need to know what version of the appliance we should use, this creates a .base_version file storing the current Anbox Cloud version the branch is targetting. This is what we do for all other repositories as well.


keirthana:
Looks good on first look, will we need a separate similar action for the gateway API yaml file?

morphis:
@keirthana We can simply extend the existing workflow to pull the gateway spec as well once it is being exported on the API.

keirthana (commented):
LGTM

PR #153, opened 2024-09-25
Automate command reference documentation

keirthana:
This PR adds a workflow to automate the generation of AMC command reference topics.

- [ ] Before merging, we need to set up a GitHub actions secret named `UBUNTU_PRO_TOKEN` with an appropriate value. We need to decide which token to use for this action.

### DIFF
--- a/.github/workflows/generate-command-reference.yml
+++ b/.github/workflows/generate-command-reference.yml
@@ -1,0 +1,24 @@
+name: Automatically generate command reference topics
+on:
+  push:
+  workflow_dispatch:
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Setup Anbox Cloud
+      uses: canonical/anbox-cloud-github-action@main
+      with:
+        ubuntu-pro-token: ${{ secrets.UBUNTU_PRO_TOKEN }}
+    - name: Checkout repo
+      uses: actions/checkout@v4
+    - name: Generate AMC command reference documentation
+      run: amc generate documentation -f markdown -d ./reference/cmd-ref/amc-command-reference
+    - name: Commit changes
+      if: ${{ github.ref == 'refs/heads/main' }}
+      uses: EndBug/add-and-commit@v9
+      with:
+        default_author: github_actions
+        message: 'Generate AMC command reference documentation'
+        add: './reference/cmd-ref/amc-command-reference'
--- a/.github/workflows/generate-command-reference.yml
+++ b/.github/workflows/generate-command-reference.yml
@@ -3,22 +3,47 @@ on:
   push:
   workflow_dispatch:
 
+env:
+  AMC_COMMAND_REFERENCE_DIR: "./reference/cmd-ref/amc"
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,29 @@
...
+
+env:
+  AMC_COMMAND_REFERENCE_DIR: "./reference/cmd-ref/amc-command-reference"
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Setup Anbox Cloud
+      uses: canonical/anbox-cloud-github-action@main
```
morphis, line 14: We should pin to a specific git revision and let renovate take care of updating when there are changes.

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,29 @@
...
+      with:
+        ubuntu-pro-token: ${{ secrets.UBUNTU_PRO_TOKEN }}
```
morphis, line 16: If you don't need to launch an instance, you can leave out the token.

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Checkout repo
+      uses: actions/checkout@v4
```
morphis, line 18: See above

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,29 @@
...
+env:
+  AMC_COMMAND_REFERENCE_DIR: "./reference/cmd-ref/amc-command-reference"
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Setup Anbox Cloud
+      uses: canonical/anbox-cloud-github-action@main
+      with:
```
morphis, line 15: We should set a channel based on what is in `.base_version`. See https://github.com/canonical/anbox-cloud-docs/blob/main/.github/workflows/update-api-specs.yaml#L22

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,29 @@
...
+        ubuntu-pro-token: ${{ secrets.UBUNTU_PRO_TOKEN }}
+    - name: Checkout repo
+      uses: actions/checkout@v4
+    - name: Delete existing files
+      run: rm -rf $AMC_COMMAND_REFERENCE_DIR/*
+    - name: Generate AMC command reference documentation
+      run: amc generate documentation -f markdown -d $AMC_COMMAND_REFERENCE_DIR
```
morphis, line 22: If we really only need `amc ` you could also do `snap install amc` instead of using the appliance which should be a lot faster.

morphis (changes requested):

.github/workflows/generate-command-reference.yml
keirthana: We need this action to do the same for appliance commands as well. So I will need the full appliance

PR #157, opened 2024-09-26
fix(gh): install missing jq tool

morphis:
yq doesn't depend on jq (it should!) so we have to explicitly install it.


keirthana (approved):
LGTM

PR #158, opened 2024-09-27
Automate command reference content

keirthana:
This PR adds a workflow to automate the generation of AMC command reference topics.

### DIFF
--- a/.github/workflows/generate-command-reference.yml
+++ b/.github/workflows/generate-command-reference.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Set $AMC_COMMAND_REFERENCE_DIR as an output
       run: echo "value=$AMC_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
       id: amc_command_reference_dir
-    - name: Commit AMC command reference changes
+    - name: Commit changes
       uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
       with:
         default_author: github_actions
@@ -37,11 +37,15 @@ jobs:
     - name: Delete existing files - appliance reference
       run: rm -rf $APPLIANCE_COMMAND_REFERENCE_DIR/*
     - name: Generate appliance command reference documentation
-      run: sudo anbox-cloud-appliance generate documentation -f markdown -d $APPLIANCE_COMMAND_REFERENCE_DIR
+      run: |
+        SNAP_DOCS_DIR="/var/snap/anbox-cloud-appliance/common/docs" && \
+        sudo mkdir -p $SNAP_DOCS_DIR && \
+        sudo anbox-cloud-appliance generate documentation -f markdown -d $SNAP_DOCS_DIR && \
+        cp -r $SNAP_DOCS_DIR/* $APPLIANCE_COMMAND_REFERENCE_DIR
     - name: Set $APPLIANCE_COMMAND_REFERENCE_DIR as an output
       run: echo "value=$APPLIANCE_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
       id: appliance_command_reference_dir
-    - name: Commit appliance command reference changes
+    - name: Commit changes
       uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
       with:
         default_author: github_actions
--- a/.github/workflows/generate-command-reference.yml
+++ b/.github/workflows/generate-command-reference.yml
@@ -3,22 +3,47 @@ on:
   push:
   workflow_dispatch:
 
+env:
+  AMC_COMMAND_REFERENCE_DIR: "./reference/cmd-ref/amc"
+  APPLIANCE_COMMAND_REFERENCE_DIR: "./reference/cmd-ref/appliance"
+
 jobs:
   run-tests:
     runs-on: ubuntu-24.04
     steps:
+    - name: Determine base version
+      id: base_version
+      run: |
+        echo "value=$(cat .base_version)" >> "$GITHUB_OUTPUT"
     - name: Setup Anbox Cloud
-      uses: canonical/anbox-cloud-github-action@main
+      uses: canonical/anbox-cloud-github-action@39490d442cc0921fe614f7044b9f3dc8bd89725c
       with:
-        ubuntu-pro-token: ${{ secrets.UBUNTU_PRO_TOKEN }}
+        channel: ${{ steps.base_version.outputs.value }}/stable
     - name: Checkout repo
-      uses: actions/checkout@v4
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+    - name: Delete existing files - amc reference
+      run: rm -rf $AMC_COMMAND_REFERENCE_DIR/*
     - name: Generate AMC command reference documentation
-      run: amc generate documentation -f markdown -d ./reference/cmd-ref/amc-command-reference
+      run: amc generate documentation -f markdown -d $AMC_COMMAND_REFERENCE_DIR
+    - name: Set $AMC_COMMAND_REFERENCE_DIR as an output
+      run: echo "value=$AMC_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
+      id: amc_command_reference_dir
     - name: Commit changes
-      if: ${{ github.ref == 'refs/heads/main' }}
-      uses: EndBug/add-and-commit@v9
+      uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
       with:
         default_author: github_actions
         message: 'Generate AMC command reference documentation'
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+      uses: canonical/anbox-cloud-github-action@39490d442cc0921fe614f7044b9f3dc8bd89725c
+      with:
+        channel: ${{ steps.base_version.outputs.value }}/stable
+    - name: Checkout repo
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+    - name: Delete existing files - amc reference
+      run: rm -rf $AMC_COMMAND_REFERENCE_DIR/*
+    - name: Generate AMC command reference documentation
+      run: amc generate documentation -f markdown -d $AMC_COMMAND_REFERENCE_DIR
+    - name: Set $AMC_COMMAND_REFERENCE_DIR as an output
```
adglkh, line 28: I am wondering why not directly use the environment variant rather than export it in one step.

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Delete existing files - appliance reference
+      run: rm -rf $APPLIANCE_COMMAND_REFERENCE_DIR/*
+    - name: Generate appliance command reference documentation
+      run: |
+        SNAP_DOCS_DIR="/var/snap/anbox-cloud-appliance/common/docs" && \
+        sudo mkdir -p $SNAP_DOCS_DIR && \
+        sudo anbox-cloud-appliance generate documentation -f markdown -d $SNAP_DOCS_DIR && \
+        cp -r $SNAP_DOCS_DIR/* $APPLIANCE_COMMAND_REFERENCE_DIR
+    - name: Set $APPLIANCE_COMMAND_REFERENCE_DIR as an output
+      run: echo "value=$APPLIANCE_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
```
adglkh, line 47: Same here

.github/workflows/generate-command-reference.yml
keirthana: It does not work when adding and committing changes if it is not exported. See line #55.

.github/workflows/generate-command-reference.yml
keirthana: Replied above

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+      uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
+      with:
+        default_author: github_actions
+        message: 'Generate AMC command reference documentation'
+        add: ${{ steps.amc_command_reference_dir.outputs.value }}
+    - name: Delete existing files - appliance reference
+      run: rm -rf $APPLIANCE_COMMAND_REFERENCE_DIR/*
+    - name: Generate appliance command reference documentation
+      run: |
+        SNAP_DOCS_DIR="/var/snap/anbox-cloud-appliance/common/docs" && \
```
jat-canonical, line 42: you can define env variables for each step as well. This can be simplified to:
```yaml
 - name: Generate appliance command reference documentation
    env:
      SNAP_DOCS_DIR: /var/snap/anbox-cloud-appliance/common/docs
    run: |
        sudo mkdir -p $SNAP_DOCS_DIR
        sudo anbox-cloud-appliance generate documentation -f markdown -d $SNAP_DOCS_DIR
        cp -r $SNAP_DOCS_DIR/* $APPLIANCE_COMMAND_REFERENCE_DIR

File: .github/workflows/generate-command-reference.yml [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+      id: base_version
+      run: |
+        echo "value=$(cat .base_version)" >> "$GITHUB_OUTPUT"
+    - name: Setup Anbox Cloud
+      uses: canonical/anbox-cloud-github-action@39490d442cc0921fe614f7044b9f3dc8bd89725c
+      with:
+        channel: ${{ steps.base_version.outputs.value }}/stable
+    - name: Checkout repo
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4
+    - name: Delete existing files - amc reference
```
jat-canonical, line 24: Since cleanup, generation and outputs are trivial steps can we combine these into one ? e.g

```yaml
    - name: Generate AMC command reference documentation
      id: amc_command_reference_dir
      run: |
         rm -rf $AMC_COMMAND_REFERENCE_DIR/*
         amc generate documentation -f markdown -d $AMC_COMMAND_REFERENCE_DIR
         echo "value=$AMC_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
```

we can do the same for appliance command reference.

jat-canonical (approved):
A couple of suggestions. Otherwise LGTM

.github/workflows/generate-command-reference.yml
adglkh: Okay, thanks. I agree with @jat-canonical put it here that folding all all those simple steps into a single one

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Set $AMC_COMMAND_REFERENCE_DIR as an output
+      run: echo "value=$AMC_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
+      id: amc_command_reference_dir
+    - name: Commit changes
+      if: ${{ github.ref == 'refs/heads/main' }}
+      uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
+      with:
+        default_author: github_actions
+        message: 'Generate AMC command reference documentation'
+        add: ${{ steps.amc_command_reference_dir.outputs.value }}
```
ajanon, line 37: How about using [`env`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#env) directly?
```suggestion
        add: ${{ env.AMC_COMMAND_REFERENCE_DIR }}
```
It seems to work from my (limited) testing.
That would remove the need to export the env variable as an output.
Wdyt?

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Delete existing files - appliance reference
+      run: rm -rf $APPLIANCE_COMMAND_REFERENCE_DIR/*
+    - name: Generate appliance command reference documentation
+      run: |
+        SNAP_DOCS_DIR="/var/snap/anbox-cloud-appliance/common/docs" && \
+        sudo mkdir -p $SNAP_DOCS_DIR && \
+        sudo anbox-cloud-appliance generate documentation -f markdown -d $SNAP_DOCS_DIR && \
+        cp -r $SNAP_DOCS_DIR/* $APPLIANCE_COMMAND_REFERENCE_DIR
```
ajanon, line 45: So, I'm guessing you have to do it in two steps because of snap confinement, right?
Would you mind adding a comment above so that it's clear why this is required?

.github/workflows/generate-command-reference.yml
```diff
@@ -0,0 +1,55 @@
...
+    - name: Set $APPLIANCE_COMMAND_REFERENCE_DIR as an output
+      run: echo "value=$APPLIANCE_COMMAND_REFERENCE_DIR" >> "$GITHUB_OUTPUT"
+      id: appliance_command_reference_dir
+    - name: Commit changes
+      if: ${{ github.ref == 'refs/heads/main' }}
+      uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
+      with:
+        default_author: github_actions
+        message: 'Generate appliance command reference documentation'
+        add: ${{ steps.appliance_command_reference_dir.outputs.value }}
```
ajanon, line 55: As above, using `env` may work:
```suggestion
        add: ${{ env.APPLIANCE_COMMAND_REFERENCE_DIR }}
```

.github/workflows/generate-command-reference.yml
ajanon, line 45: If you want to ensure that each line only gets run if the previous one succeeded, how about using `set -e`?
```suggestion
        set -e
        sudo mkdir -p $SNAP_DOCS_DIR
        sudo anbox-cloud-appliance generate documentation -f markdown -d $SNAP_DOCS_DIR
        cp -r $SNAP_DOCS_DIR/* $APPLIANCE_COMMAND_REFERENCE_DIR
```
From the [bash manual](https://manpages.ubuntu.com/manpages/noble/en/man1/bash.1.html):
> -e      Exit  immediately  if  a  pipeline  (which  may  consist of a single simple
                      command), a list, or a compound command (see SHELL  GRAMMAR  above),  exits
                      with  a non-zero status.

ajanon (changes requested):
Looks pretty good!
Some comments from my end.

ajanon (approved):
LGTM, thanks!

PR #160, opened 2024-09-27
Automate generating the AMS configuration attributes

keirthana:
This PR contains updates that will automatically generate the AMS configuration attributes from the AMS API swagger.

Currently, the AMS configuration reference topic is pointing `ams-configuration-public.md` which is manually created.
For 1.23.2, the swagger output will be refreshed and the `ams-configuration.md` will be updated. Once that is verified, the topic needs to be switched in the TOC to point to the automatically generated `ams-configuration.md`.


jat-canonical (approved):
If the build failures can be ignored LGTM

keirthana:
> If the build failures can be ignored LGTM

Yes, those are not specific to this PR. I will fix them separately.

PR #165, opened 2024-10-04
fix(gh): use correct yq option on noble

morphis:
Yq on jammy has -P for emitting YAMl from JSON but on noble a newer verison of yq used -y for that.

-----

This is the last fix necessary to get the workflow to actually do its job :-)


keirthana (approved):
LGTM

PR #166, opened 2024-10-07
Update workflow to create a PR instead of direct push

keirthana:
The current workflow lacks permission to commit to main directly. So modifying the workflow to create a PR instead.


.github/workflows/generate-command-reference.yml
```diff
@@ -15,41 +15,39 @@ jobs:
...
-        add: ${{ env.APPLIANCE_COMMAND_REFERENCE_DIR }}
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        commit-message: "Commit changes to command reference documentation"
+        branch-name: update-cmd-ref
+        title: Update command reference documentation
+        body: |
+          This updates the command reference topics for the AMC and Anbox Cloud Appliance.
+        repository: canonical/anbox-cloud-docs
+        upsert: true
+        ignore-no-changes: true
```
ajanon, line 53: Oh, this looks much nicer to use than the base gh command line tool!

ajanon (approved):
LGTM

PR #181, opened 2024-10-09
Exempt auto-generated and venv files from spellcheck

keirthana:
Because of the way the current outputs are generated for AMS configuration and API reference the spellcheck is flagging terminology from files inside the venv and the automatically generated command reference files.

Until we change the venv directory and fix our CLI help text, we need this exemption in place so that the spellcheck does not fail on other unrelated changes.


ajanon (approved):
LGTM

PR #182, opened 2024-10-09
Update messaging around custom images

keirthana:


ajanon (approved):
LGTM.
Just a heads-up though: it will probably be supported for Android 14 AAOS in the future.

PR #184, opened 2024-10-09
Publish auto-generated command reference topics

keirthana:
This PR will make the auto-generated command reference files visible in public documentation. 

The current auto-generated files can be further optimised but that work is planned for later. Right now, since the information is accurate enough, we can publish the auto-generated files.


ajanon (approved):
LGTM

PR #190, opened 2024-10-10
Add Android 14 images to the supported images list for 1.24

keirthana:
Adding only AOSP Android 14 images to the list for 1.24.0. AAOS, if supported, will have to be added for 1.24.1.

Decision on [PackageManager](https://developer.android.com/reference/android/content/pm/PackageManager) level feature listing is that we should consider providing a list of features but we will need to set it automatically (maybe using [feature list generator](https://github.com/canonical/anbox/blob/main/src/anbox/android/features_list_generator.cpp)). This needs more research and hence not covered in this PR.


ajanon (approved):
LGTM, thanks!

adglkh (approved):
LGTM

PR #197, opened 2024-10-17
Fix ignore list for spellcheck

keirthana:
Words that could be used both in lower case and capital case should be added in lower case to the spellcheck ignore list. The spell check automatically considers proper forms of capitalisation.

For example: If `ams` is added, `ams`, `Ams` and `AMS` will be ignored but `aMs` won't be ignored.

This PR updates some of the words that were added as upper case but has lower case uses in text.


adglkh (approved):
LGTM

PR #203, opened 2024-10-25
Add more features to the unsupported list for AAOS

ajanon:
Some features were not listed as unsupported for AAOS, but they are. Fix this by adding them to the table and explaining how it works for AAOS.


morphis (approved):
LGTM, thanks!

keirthana (approved):
LGTM

PR #205, opened 2024-11-01
Add streaming tutorial

keirthana:
This PR adds the stream client tutorial to the documentation.

### DIFF
--- a/tutorial/stream-client.md
+++ b/tutorial/stream-client.md
@@ -21,7 +21,7 @@ To access the HTTP API of the Anbox Cloud stream gateway, an access token is req
 
 On the machine where Anbox Cloud Appliance is installed, create the service account by running the following command:
 
-    sudo anbox-cloud-appliance gateway account create streaming-tutorial
+    sudo anbox-cloud-appliance.gateway account create streaming-tutorial
 
 The output of this command provides the access token. Make a note of this token to use when you make a request to the stream gateway API.
 
@@ -107,7 +107,7 @@ gateway_server_cert = os.getenv("GATEWAY_SERVER_CERT") or False
 gateway_enabled = bool(gateway_api_url and gateway_api_token)
 
 auth_user = os.getenv("AUTH_USER") or "anbox"
-auth_password = os.getenv("AUTH_PASSOWRD") or secrets.token_urlsafe(16)
+auth_password = os.getenv("AUTH_PASSWORD") or secrets.token_urlsafe(16)
 
 print(f"You can login with username {auth_user} and password {auth_password}")
 
--- a/tutorial/stream-client.md
+++ b/tutorial/stream-client.md
@@ -1,0 +1,259 @@
+This tutorial guides you through the process of setting up a web-based streaming client using the Anbox Cloud streaming stack. The connection between the stream client and the server uses WebRTC backed by web sockets which enable the real time communications required for streaming. To know more about the WebRTC configuration, see [WebRTC streamer](https://discourse.ubuntu.com/t/webrtc-streamer/30195).
+
+TBD add more context from https://discourse.ubuntu.com/t/issues-with-anbox-streaming-sdk/47778
+
+## Preparation
+
+Complete the following preparatory steps:
+
+### Install the Anbox Cloud Appliance
+
+We need the Anbox Cloud streaming stack to be deployed already to set up a streaming client. So let's get the streaming stack ready by installing the Anbox Cloud Appliance. Follow the instructions in the [appliance installation tutorial](https://discourse.ubuntu.com/t/22681) until you finish initialising the Appliance.
+
+### Create an access token
+
+To access the HTTP API of the Anbox Cloud stream gateway, an access token is required. Each access token is associated with a service account.
+
+On the machine where Anbox Cloud Appliance is installed, create the service account by running the following command:
+
+    anbox-cloud-appliance gateway account create test
+
+The output of this command provides the access token. Make a note of this token to use when you make a request to the stream gateway API.
+
+See [How to access the stream gateway](https://discourse.ubuntu.com/t/how-to-access-the-stream-gateway/17784) for more information on creating, using and deleting the access token.
+
+### Create an Android Instance
+
+Create a streaming enabled Android instance:
+
+    amc launch --enable-streaming --raw --name a13 jammy:android13:amd64
+
+### Determine session ID of the Android instance
+
+For the next step we need the session ID which AMS assigned to the instance:
+
+    amc ls --filter name=a13 --format=csv | cut -d, -f6 | awk -F"session=" '{ print $2 }'
+
+## Implement the stream client
+
+Create a directory to set up the stream client:
+
+    mkdir -p stream-client/{templates,static}
+
+Create a `requirements.txt` file inside `stream-client`
+
+```
+flask==3.0.3
+requests==2.32.3
+```
+
+Create a `service.py` file inside `stream-client`:
+
+```python
+import os
+import os.path
+import secrets
+
+from flask import Flask, jsonify, render_template, request
+from requests import Session
+from functools import wraps
+from typing import *
+
+
+class GatewayAPI:
+    def __init__(self, base_url, session, token, verify):
+        self.base_url = base_url
+        self.session = session
+        self.token = token
+        self.verify = verify
+
+    def make_request(
+        self,
+        method,
+        path,
+        headers,
+        data,
+    ):
+        uri = f"{self.base_url}{path}"
+
+        headers["Authorization"] = f"macaroon root={self.token}"
+        headers["Content-Type"] = "application/json"
+
+        return self.session.request(
+            method, uri, json=data, headers=headers, verify=self.verify
+        )
+
+    def join_session(self, session_id, data):
+        uri = f"/1.0/sessions/{session_id}/join"
+        return self.make_request(
+            "POST",
+            uri,
+            data=data,
+            headers={"Content-Type": "application/json"},
+        ).json()
+
+
+app = Flask(__name__)
+
+gateway_api_url = os.getenv("GATEWAY_URL")
+gateway_api_token = os.getenv("GATEWAY_API_TOKEN")
+gateway_server_cert = os.getenv("GATEWAY_SERVER_CERT") or False
+gateway_enabled = bool(gateway_api_url and gateway_api_token)
+
+auth_user = os.getenv("AUTH_USER") or "anbox"
+auth_password = os.getenv("AUTH_PASSOWRD") or secrets.token_urlsafe(16)
+
+print(f"You can login with username {auth_user} and password {auth_password}")
+
+gateway = GatewayAPI(gateway_api_url, Session(), gateway_api_token, gateway_server_cert)
+
+
+def render_error_response(msg, code):
+    resp = {"error_msg": msg}
+    return jsonify(resp), code
+
+
+def check_auth(username, password):
+    return username == auth_user and password == auth_password
+
+
+def login_required(f):
+    @wraps(f)
+    def wrapped_view(**kwargs):
+        auth = request.authorization
+        if not (auth and check_auth(auth.username, auth.password)):
+            return (
+                "Unauthorized",
+                401,
+                {"WWW-Authenticate": 'Basic realm="Login Required"'},
+            )
+        return f(**kwargs)
+
+    return wrapped_view
+
+
+@app.route("/1.0/sessions/<session_id>/join", methods=["POST"])
+@login_required
+def api_1_0_sessions_join_post(session_id):
+    if not gateway_enabled:
+        return render_error_response("no gateway connected", 503)
+
+    req = {"disconnect_clients": True}
+    resp = gateway.join_session(session_id, req)
+
+    if "status_code" not in resp or resp["status_code"] != 200:
+        return render_error_response("failed to join session", 500)
+
+    return jsonify(resp), 200
+
+
+@app.route("/<session_id>")
+@login_required
+def stream_session(session_id=None):
+    return render_template("stream.html", session_id=session_id)
+
+
+@app.route("/")
+@login_required
+def index():
+    return "Welcome to the Anbox Cloud Stream SDK tutorial!"
+```
+
+Create a `stream.html` file inside `stream-client/static`:
+
+```html
+<!doctype html>
+<html>
+  <head>
+    <title>Anbox Streaming SDK - Tutorial</title>
+    <script type="module" src="static/anbox-stream-sdk.js"></script>
+    <style>
+      html,
+      body,
+      #anbox-stream {
+        height: 100%;
+        margin: 0;
+        width: 100%;
+      }
+    </style>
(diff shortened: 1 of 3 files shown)


tutorial/stream-client.md
```diff
@@ -0,0 +1,260 @@
...
+When installing and initialising the appliance, we would have configured the private IP address of the machine running the appliance. The gateway API is accessible on the local listen address of the appliance on port `9031/tcp`. So that's the IP address and port we use in the GATEWAY_URL.
+
+    export GATEWAY_URL=https://1.2.3.4:9031
+    export GATEWAY_API_TOKEN=xxx
+    export GATEWAY_SERVER_CERT="$PWD"/gateway.crt
+    FLASK_APP=service.py python3 -m flask run -h 1.2.3.4 -p 8080
+
+This will start the service and make it accessible at http://1.2.3.4:8080 and print out a username and password needed to access the site.
+
+To access the stream of the instance we created earlier, you can access `http://1.2.3.4:8080/<session_id>` in your browser. Remember to use the session ID of the Android instance we retrieved earlier in place of `<session_id>`.
```
morphis, line 260: We should give a hint that they can login via `anbox` and the password the example has printed.

tutorial/stream-client.md
```diff
@@ -0,0 +1,260 @@
...
+
+Now that we have everything ready, let's create a directory to set up the stream client:
+
+    mkdir -p stream-client/{templates,static}
+
+Create a `requirements.txt` file inside `stream-client` with the following code:
+
+```
+flask==3.0.3
+requests==2.32.3
```
morphis, line 51: Can you add `pyopenssl=24.2.1`?

tutorial/stream-client.md
```diff
@@ -0,0 +1,260 @@
...
+    pip3 install -r requirements.txt
+
+To run the stream client service application, we need to set the necessary environment variables and execute the service via flask.
+
+When installing and initialising the appliance, we would have configured the private IP address of the machine running the appliance. The gateway API is accessible on the local listen address of the appliance on port `9031/tcp`. So that's the IP address and port we use in the GATEWAY_URL.
+
+    export GATEWAY_URL=https://1.2.3.4:9031
+    export GATEWAY_API_TOKEN=xxx
+    export GATEWAY_SERVER_CERT="$PWD"/gateway.crt
+    FLASK_APP=service.py python3 -m flask run -h 1.2.3.4 -p 8080
```
morphis, line 256:
```suggestion
    FLASK_APP=service.py python3 -m flask --cert=adhoc run -h 1.2.3.4 -p 8080
```

This will give us TLS with an adhoc generated certificate the user has to accept in the browser. Not meant to be used in production but makes us use TLS always.

tutorial/stream-client.md
```diff
@@ -0,0 +1,260 @@
...
+
+This will start the service and make it accessible at http://1.2.3.4:8080 and print out a username and password needed to access the site.
```
morphis, line 258:
```suggestion
This will start the service and make it accessible at https://1.2.3.4:8080 and print out a username and password needed to access the site.
```

morphis (changes requested):

adglkh:
@morphis 
Since we want people to implement their stream client by talking to API service which is implemented on their own as a good practice, for all examples under the anbox-streaming-sdk, what's the plan for this? 
We leave them as they are, meaning they remain directly talking to the stream gateway endpoint via the 9031 port or we change this at some later date    
https://github.com/canonical/anbox-streaming-sdk/blob/main/examples/js/anime4K.html#L23
https://github.com/canonical/anbox-streaming-sdk/blob/main/examples/js/example.html#L23
https://github.com/canonical/anbox-streaming-sdk/blob/main/examples/android/enhanced_webview_streaming/app/src/main/assets/index.html#L24

morphis:
> Since we want people to implement their stream client by talking to API service which is implemented on their own as a good practice, for all examples under the anbox-streaming-sdk, what's the plan for this?
> We leave them as they are, meaning they remain directly talking to the stream gateway endpoint via the 9031 port or we change this at some later date

I would like to get the examples changed in a next step and to be converted to a more gallery like type of application to allow demonstrating different aspects. This should then have a similar flask based service in the background.

tutorial/stream-client.md
keirthana: Just to clarify: `--cert=adhoc` is the part that is not meant for production right? Otherwise, this entire example of setting up a stream client is something that users can use in production?

tutorial/stream-client.md
```diff
@@ -0,0 +1,261 @@
...
+    </script>
+  </body>
+</html>
+```
+
+Next, let's make a copy of the Anbox Stream Gateway server to allow the service to validate it:
+
+    # We use cat here to ensure gateway.crt has the users permissions
+    sudo cat /var/snap/anbox-cloud-appliance/common/gateway/server.crt > gateway.crt
+
```
ajanon, line 232: Should users do this in general, or should it be limited to small scale tests?
I feel like this shouldn't be done in production, but I'm not too sure if it's a problem.
@adglkh @morphis @jat-canonical Wdyt?

tutorial/stream-client.md
```diff
@@ -0,0 +1,261 @@
...
+Now, we can use `pip` to install all required dependencies
+
+    pip3 install -r requirements.txt
+
+To run the stream client service application, we need to set the necessary environment variables and execute the service via flask.
+
+When installing and initialising the appliance, we would have configured the private IP address of the machine running the appliance. The gateway API is accessible on the local listen address of the appliance on port `9031/tcp`. So that's the IP address and port we use in the GATEWAY_URL.
+
+    export GATEWAY_URL=https://1.2.3.4:9031
+    export GATEWAY_API_TOKEN=xxx
```
ajanon, line 255: But what token should be input by the user here? Should we be explicit about how to do this? Or maybe link them to some other docs?

ajanon (commented):
This is pretty neat! I have two small questions.

tutorial/stream-client.md
morphis: No, using a flask server like this is not meant to be used in production. We can for example to link to https://flask.palletsprojects.com/en/stable/deploying/

tutorial/stream-client.md
morphis: @ajanon what specifically do you mean?

tutorial/stream-client.md
morphis: We have that above under prerequisites but we should put to that in the description here

tutorial/stream-client.md
morphis: No, it's not about using the same certificate (we would still miss the private key for that). This is simply to make the flask service add the cert to its list of trusted certs so it can validate the connection with the gateway service at the TLS level. Does that answer your question?

tutorial/stream-client.md
lorumic: I agree, I propose to chain the output of the command above here, so that it is used directly in the context where it is needed:
```suggestion
    export GATEWAY_API_TOKEN=$(anbox-cloud-appliance gateway account create your-username)
```

tutorial/stream-client.md
```diff
@@ -0,0 +1,261 @@
...
+
+Now, we can use `pip` to install all required dependencies
+
+    pip3 install -r requirements.txt
+
+To run the stream client service application, we need to set the necessary environment variables and execute the service via flask.
+
+When installing and initialising the appliance, we would have configured the private IP address of the machine running the appliance. The gateway API is accessible on the local listen address of the appliance on port `9031/tcp`. So that's the IP address and port we use in the GATEWAY_URL.
+
+    export GATEWAY_URL=https://1.2.3.4:9031
```
lorumic, line 254: Maybe replace 1.2.3.4 with `<your-private-ip>`?

tutorial/stream-client.md
```diff
@@ -0,0 +1,261 @@
...
+    export GATEWAY_API_TOKEN=xxx
+    export GATEWAY_SERVER_CERT="$PWD"/gateway.crt
+    FLASK_APP=service.py python3 -m flask --cert=adhoc run -h 1.2.3.4 -p 8080
+
+This will start the service and make it accessible at https://1.2.3.4:8080 and print out a username and password needed to access the site.
+
+To access the stream of the instance we created earlier, you can access `https://1.2.3.4:8080/<session_id>` in your browser. Remember to use the session ID of the Android instance we retrieved earlier in place of `<session_id>` and for credentials, use `anbox` as username and the password that the example printed.
```
lorumic, line 261: I would never use this 1.2.3.4. It can be confusing. Better specify that this is the private IP address of the machine explicitly, just like it is explicitly specified that one should use the `<session_id>` retrieved above.

Also, "access" is repeated twice. Better use "open" the second time:

```suggestion
    FLASK_APP=service.py python3 -m flask --cert=adhoc run -h 1.2.3.4 -p 8080

This will start the service and make it accessible at https://1.2.3.4:8080 and print out a username and password needed to access the site.

To access the stream of the instance we created earlier, you can open `https://1.2.3.4:8080/<session_id>` in your browser. Remember to use the session ID of the Android instance we retrieved earlier in place of `<session_id>` and for credentials, use `anbox` as username and the password that the example printed.
```

tutorial/stream-client.md
keirthana: We are going through a learning flow in the tutorial where we help the user prepare everything they need and then when they use it later, it helps them understand how various things work together.

So I have chosen to add text description about the token rather than chaining the commands

morphis (changes requested):
I did a trial run of this on a local VM and found a few things which were not yet working

@canonical/anbox once the changes are merged, does anyone volunteer to give this another trial run?

tutorial/stream-client.md
```diff
@@ -0,0 +1,267 @@
...
+
+    curl -o stream-client/static/anbox-stream-sdk.js \
+        https://raw.githubusercontent.com/canonical/anbox-streaming-sdk/refs/heads/main/js/anbox-stream-sdk.js
+
+## Run the stream client
+
+To run the stream client, let's set up a virtual environment for Python to install necessary dependencies:
+
+    cd stream-client
+    python3 -m venv .venv
```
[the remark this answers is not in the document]
keirthana: I have rephrased the flow slightly differently to consider this scenario

tutorial/stream-client.md
```diff
@@ -0,0 +1,267 @@
...
+    export GATEWAY_SERVER_CERT="$PWD"/gateway.crt
+    FLASK_APP=service.py python3 -m flask --cert=adhoc run -h <appliance_private_ip> -p 8080
+
+This will start the service and make it accessible at `https://<appliance_private_ip>:8080` and print out a username and password needed to access the site.
+
+```{note}
+This example uses a Flask server with an adhoc certificate. Such an implementation is not recommended for production deployments. For better ways to deploy to production, see the [Flask documentation](https://flask.palletsprojects.com/en/stable/deploying/).
+```
+
+To view the stream of the instance we created earlier, you can access `https://<appliance_private_ip>:8080/<session_id>` in your browser. Remember to use the session ID of the Android instance we retrieved earlier in place of `<session_id>` and for credentials, use `anbox` as username and the password that the example printed.
```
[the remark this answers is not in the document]
keirthana: Yes, I have included that now

tutorial/stream-client.md
```diff
@@ -0,0 +1,276 @@
...
+We need the Anbox Cloud streaming stack to be deployed already to set up a streaming client. So let's get the streaming stack ready by installing the Anbox Cloud Appliance. Follow the instructions in the {ref}`tut-installing-appliance` until initialising the appliance. When you have installed and initialised the appliance, let's proceed to the next step.
+
+(sec-create-access-token)=
+### Create an access token
+
+To access the HTTP API of the Anbox Cloud stream gateway, an access token is required. Each access token is associated with a service account.
+
+On the machine where Anbox Cloud Appliance is installed, create the service account by running the following command:
+
+    sudo anbox-cloud-appliance gateway account create streaming-tutorial
```
adglkh, line 24: ->
`sudo anbox-cloud-appliance.gateway account create streaming-tutorial`
I don't think we provide the subcommand `gateway` to the anbox-cloud-appliance snap

tutorial/stream-client.md
```diff
@@ -0,0 +1,276 @@
...
+
+app = Flask(__name__)
+
+gateway_api_url = os.getenv("GATEWAY_URL")
+gateway_api_token = os.getenv("GATEWAY_API_TOKEN")
+gateway_server_cert = os.getenv("GATEWAY_SERVER_CERT") or False
+gateway_enabled = bool(gateway_api_url and gateway_api_token)
+
+auth_user = os.getenv("AUTH_USER") or "anbox"
+auth_password = os.getenv("AUTH_PASSOWRD") or secrets.token_urlsafe(16)
```
adglkh, line 110: AUTH_PASSOWRD
->
AUTH_PASSWORD

tutorial/stream-client.md
```diff
@@ -0,0 +1,276 @@
...
+            method, uri, json=data, headers=headers, verify=self.verify
+        )
+
+    def join_session(self, session_id, data):
+        uri = f"/1.0/sessions/{session_id}/join"
+        return self.make_request(
+            "POST",
+            uri,
+            data=data,
+            headers={"Content-Type": "application/json"},
```
adglkh, line 99: Is this needed? otherwise you will override the `Content-Type` header in the make_request function.

tutorial/stream-client.md
```diff
@@ -0,0 +1,276 @@
...
+@login_required
+def api_1_0_sessions_join_post(session_id):
+    if not gateway_enabled:
+        return render_error_response("no gateway connected", 503)
+
+    request_data = {"disconnect_clients": True}
+    response = gateway.join_session(session_id, request_data)
+
+    if "status_code" not in response or response["status_code"] != 200:
+        return render_error_response("failed to join session", 500)
```
adglkh, line 151: Maybe we want to include the error message in the response as well?
https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/api-reference/gateway-api/#/session/handle-join-session

tutorial/stream-client.md
keirthana: Great catch, thanks!

tutorial/stream-client.md
keirthana: @adglkh You mean to say that the header declaration line is unnecessary or that it will cause errors?

tutorial/stream-client.md
keirthana: Agreed, that should be added to the API documentation in swagger.

morphis: I would like to do some more refinements on the tutorial once the first version is out where we can include such things. How does that sound @adglkh ?

adglkh: Sound good to me!

adglkh (approved):
LGTM

jat-canonical (approved):
LGTM

morphis:
Only link check is failing, so fine to merge.

PR #209, opened 2024-11-06
feat(explanation): introduce Anbox VHAL adpater

adglkh:


explanation/aaos.md
```diff
@@ -9,6 +9,12 @@ Since the 1.21.0 release, in addition to providing images based on the Android O
 
 The [Vehicle Abstraction Layer (VHAL)](https://source.android.com/docs/automotive/vhal) enables Android Automotive to communicate with and control the different vehicle hardware controls.
 
+Before the 1.24.0 release, if you created your own VHAL implementation by following the [instructions](https://documentation.ubuntu.com/anbox-cloud/en/latest/howto/android/custom-vhal/) and loaded it as an addon during the Android runtime, vehicle properties and configurations were not visible on the Anbox Cloud dashboard. With the 1.24.0 release, Anbox Cloud introduced a VHAL adapter that acts as a bridge between the Anbox runtime and the VHAL implementation for data transmission. The Anbox VHAL adapter processes messages sent from the Anbox runtime, performs operations against the VHAL implementation(such as retrieving a list of property configurations), and sends the results back to the Anbox runtime. As a result, vehicle properties and configurations are now visible on the Anbox Cloud dashboard for the third-party VHAL implementations.
```
keirthana, line 12:
```suggestion
Until the 1.24.0 release, if you created your own VHAL implementation by following the [our instructions for customising the Anbox VHAL](https://documentation.ubuntu.com/anbox-cloud/en/latest/howto/android/custom-vhal/) and loaded it as an addon during the Android runtime, vehicle properties and configurations were not visible on the Anbox Cloud dashboard. With the 1.24.0 release, Anbox Cloud has introduced a VHAL adapter that acts as a bridge between the Anbox runtime and the VHAL implementation for data transmission. 

The Anbox VHAL adapter processes messages sent from the Anbox runtime, performs operations against the VHAL implementation (such as retrieving a list of property configurations), and sends the results back to the Anbox runtime. As a result, vehicle properties and configurations are now visible on the Anbox Cloud dashboard for third-party VHAL implementations.
```

explanation/aaos.md
```diff
@@ -9,6 +9,12 @@ Since the 1.21.0 release, in addition to providing images based on the Android O
...
+
+If the third-party VHAL implementation is loaded during Android runtime, on the Anbox Cloud Dashboard, some vehicle property values may not be accessed, while others may not be editable due to [permission controls](https://source.android.com/docs/automotive/vhal/previous/properties#vehicle-props) that categorise vehicle properties as read-only, write-only, or read-write.
```
keirthana, line 14:
```suggestion
However, if a third-party VHAL implementation is loaded during Android runtime, some vehicle property values may still not be accessible while others may not be editable, on the Anbox Cloud dashboard, due to [permission controls](https://source.android.com/docs/automotive/vhal/previous/properties#vehicle-props) that categorise vehicle properties as read-only, write-only, or read-write.
```

keirthana:
@morphis @adglkh A more general question: This feels like we are actually releasing the full ability to be able to access/modify the properties in 1.24.1. So why don't we actually release this documentation as well, maybe a bit rephrased to remove the mentions of 1.24, along with 1.24.1?

Because it seems to me like we are telling the users we have done this but they cannot fully use it until 1.24.1.

keirthana (changes requested):

adglkh:
> @morphis @adglkh A more general question: This feels like we are actually releasing the full ability to be able to access/modify the properties in 1.24.1. So why don't we actually release this documentation as well, maybe a bit rephrased to remove the mentions of 1.24, along with 1.24.1?
> 
> Because it seems to me like we are telling the users we have done this but they cannot fully use it until 1.24.1.
> 
> Also, if we decide to keep this in 1.24.0, I will work with you @adglkh about the placement of this content, I feel it would be better to place in the how-to guide for customising VHAL for better context.

A: Let me clarify, as there seems to be some confusion about this functionality.

With the Anbox VHAL implementation and VHAL adapter included in the 1.24.0 LXD image, users can:

- access non-readable vehicle property values
- modify non-writable vehicle property values

directly from the dashboard without issues.

However, this capability does not extend to third-party VHAL implementations. To gain the same access and control as the Anbox VHAL implementation, third-party implementations must implement the Anbox-specific HIDL interfaces we’ve defined. Currently, the HIDL interfaces are closed-source, but after discussions with the team, we plan to release them as open-source repo on GitHub[AC-2821]. This will enable third-party VHAL implementations to leverage these interfaces and access or modify vehicle property values in the same way as the Anbox VHAL implementation.

keirthana:
> > @morphis @adglkh A more general question: This feels like we are actually releasing the full ability to be able to access/modify the properties in 1.24.1. So why don't we actually release this documentation as well, maybe a bit rephrased to remove the mentions of 1.24, along with 1.24.1?
> > Because it seems to me like we are telling the users we have done this but they cannot fully use it until 1.24.1.
> > Also, if we decide to keep this in 1.24.0, I will work with you @adglkh about the placement of this content, I feel it would be better to place in the how-to guide for customising VHAL for better context.
> 
> A: Let me clarify, as there seems to be some confusion about this functionality.
> 
> With the Anbox VHAL implementation and VHAL adapter included in the 1.24.0 LXD image, users can:
> 
>     * access non-readable vehicle property values
> 
>     * modify non-writable vehicle property values
> 
> 
> directly from the dashboard without issues.
> 
> However, this capability does not extend to third-party VHAL implementations. To gain the same access and control as the Anbox VHAL implementation, third-party implementations must implement the Anbox-specific HIDL interfaces we’ve defined. Currently, the HIDL interfaces are closed-source, but after discussions with the team, we plan to release them as open-source repo on GitHub[AC-2821]. This will enable third-party VHAL implementations to leverage these interfaces and access or modify vehicle property values in the same way as the Anbox VHAL implementation.

Thanks for clarifying!

explanation/aaos.md
adglkh: Fixed as suggested.
Thanks

keirthana (approved):
Thanks, LGTM

ajanon (approved):
LGTM

Issue #206, opened 2024-11-05
Review juju crashdump command


lathiat:
While trying to use juju-crashdump, I found the following

1. With juju 3.5.x, you seem to need to use `juju-crashdump` rather than `juju crashdump`, it seems the plugin support was maybe dropped or is broken. The juju-crashdump docs don't currently reflect this.
2. With the command given, it says that as_root is needed, so perhaps we need to add --as-root to the command?
    `2024-11-05 06:32:56,790 - The as_root flag must be used to run addon debug-layer`

---
Document: howto/troubleshoot/landing.md

adglkh:
Thanks for the report.
Fixed it in the following PR
https://github.com/canonical/anbox-cloud-docs/pull/218

PR #211, opened 2024-11-06
Fix juju crashdump command

keirthana:
To fix issue #206

@canonical/anbox I validated this in theory but not sure if I can test this command in Appliance because it does not apply to the Appliance version of our deployment. Could one of you confirm or validate?


adglkh (approved):
LGTM

PR #212, opened 2024-11-06
Create 1.24.0 release documentation

keirthana:
* Add release notes for 1.24.0
* Update release information
* Add component versions
* Remove 20.04 information

Pending:

- [x] Update component versions after candidate is ready
- [x] Confirm removal of node controller information across documentation
- [x] Add dashboard improvements section from David's team
- [x] Inline comments seeking clarification
- [x] Known issue for how to force [displaying an activity](https://github.com/canonical/anbox-streaming-sdk/blob/main/js/anbox-stream-sdk.js#L99) via the streaming SDK when using Android 14 image
- [x] Update bundle versions after they are promoted to stable

### DIFF
--- a/reference/release-notes/1.24.0.md
+++ b/reference/release-notes/1.24.0.md
@@ -1,0 +1,96 @@
+---
+orphan: true
+---
+# 1.24.0
+
+These release notes cover new features and changes in Anbox Cloud 1.24.0.
+
+Anbox Cloud 1.24.0 is a minor release. To understand minor and patch releases, see [Release notes](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/release-notes/release-notes).
+
+Please see [Component versions](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/component-versions/) for a list of updated components.
+
+## Requirements
+
+See the [Requirements](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/requirements/) for details on general and deployment specific requirements to run Anbox Cloud.
+
+## New features & improvements
+
+### Updates to components
+
+* Anbox Cloud includes and uses the following latest versions for components it uses:
+  - LXC version 6.0.2<!--AC-2812-->
+  - Mesa driver version 24.2.6<!--AC-2800-->
+  - VirGL renderer version 1.9.0.<!--AC-2800-->
+
+### Security
+
+* We have enhanced security with a minimum version of TLS 1.3 as a requirement for all services working with Anbox Cloud. TLS 1.3 is also enabled by default for the Anbox Stream Agent and Gateway.<!--AC-2728 and AC-2654-->
+
+### Anbox Management Service (AMS)
+* We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
+* Anbox Cloud provides the image type (`android`, `aaos` or `generic`) through the following three ways:<!--AC-2757-->
+  - `metadata.yaml` of the Anbox Cloud image
+  - API response from the AMS when querying image information
+  - Output of `amc image show` command
+
+  This allows the web client to adapt its interface according to the image type.
+* We have enabled support for multiple cluster management by AMS.<!--AC-2674-->
+
+### Android
+
+* This release includes Android 14 based AOSP images.
+* We now support using the lock screen in Anbox Cloud([LP 1983426](https://bugs.launchpad.net/anbox-cloud/+bug/1983426)).
+
+### Charms
+
+* For the NFS Operator charm for Anbox Cloud, `Cachefilesd` is now optional and is used only when `fsc` is provided as a mount option.<!--AC-2839-->
+
+### Streaming improvements
+
+* Anbox Cloud now provides an option to allow the use of custom fragment shader-based upscaling algorithms for video streaming. This helps you to apply multiple shaders sequentially to perform multi-pass upscaling and replaces the default [AMD FidelityFX Super Resolution 1.0](https://gpuopen.com/fidelityfx-superresolution/) shader. See {ref}`sec-custom-fragment-shader` for more information.<!--AC-2541-->
+* The Anbox Streaming SDK now uses the [Snapdragon™ Game Super Resolution(SGSR)](https://github.com/SnapdragonStudios/snapdragon-gsr) algorithm, which integrates upscaling and sharpening in one single GPU shader pass, by default.<!--AC-2543-->
+
+### Other
+
+* Android security updates for November 2024 (see [Android Security Bulletin - November 2024](https://source.android.com/docs/security/bulletin/2024-11-01) for more information).
+* The Android WebView has been updated to 130.0.6723.73.
+
+## Removed functionality
+
+Following deprecation in earlier releases, these functionalities have been removed from Anbox Cloud with the 1.24.0 release:
+
+* Support for Ubuntu 20.04 (Focal Fossa)
+* Use of node controller to manage the port forwarding from an instance to expose a service
+
+## Deprecations
+
+< List deprecated features. Provide an alternative, if any, for the users to use instead of the deprecated features >
+
+## Known issues
+
+* Out of band support does not work as expected for Android 14 images because of a change in implicit intents feature in the upstream.<!--AC-2852-->
+* The following features are supported for AAOS images based on Android 14. This is not an intended support and will be removed in the next patch release:<!--AC-2848-->
+  - `enable_virtual_keyboard`
+  - `enable_system_ui`
+  - `enable_anbox_ime`
+  - Custom Android ID
+  - Custom system apps
+  - Boot package support
+
+## Bug fixes
+
+The following bugs have been fixed as part of the Anbox Cloud 1.24.0 release.
+
+* [LP 2064900](https://bugs.launchpad.net/anbox-cloud/+bug/2064900) If an instance fails and ends up with *Error* status, the session for the instance is set to *Error* status as well. When the instance is restarted, the session status remains in *Error* and it is not possible to stream from the instance again.
+* [LP 2084897](https://bugs.launchpad.net/anbox-cloud/+bug/2084897) Null pointer exception when opening Anbox IME.
+* [LP 2072496](https://bugs.launchpad.net/anbox-cloud/+bug/2072496) Setting user data while creating an instance with streaming enabled creates an instance but does not allow you to join the session.
+* [LP 2085098](https://bugs.launchpad.net/anbox-cloud/+bug/2085098) The dashboard fails to connect to the stream gateway because the value of the `ASG_API_URL` configuration is used to limit the dashboard's communication. So when the CORS header is set to, for example, the private address of the gateway, the dashboard cannot connect to the gateway.
+* [LP 2084800](https://bugs.launchpad.net/anbox-cloud/+bug/2084800) The `no_wait` query parameter for the `POST /1.0/instances` endpoint is not documented.
+* [LP 2084120](https://bugs.launchpad.net/anbox-cloud/+bug/2084120) Anbox WebRTC Data Proxy service takes too long to get started.
+* [LP 2083795](https://bugs.launchpad.net/anbox-cloud/+bug/2083795) When reconnecting to the ADB server, the connection is refused.
+* [LP 2077638](https://bugs.launchpad.net/anbox-cloud/+bug/2077638) After initialising the Anbox Cloud Appliance, the output of `amc image list` remains empty. This shows that the Anbox Management Service (AMS) has not loaded the image list immediately after the initialisation.
+* [LP 2084253](https://bugs.launchpad.net/anbox-cloud/+bug/2084253) 64-bit only detection on ARM systems does not work and causes a boot failure.
+
+## Upgrade instructions
+
+See [How to upgrade Anbox Cloud](https://documentation.ubuntu.com/anbox-cloud/en/latest/howto/update/upgrade-anbox/#howto-upgrade-anbox-cloud) and [How to upgrade the Anbox Cloud Appliance](https://documentation.ubuntu.com/anbox-cloud/en/latest/howto/update/upgrade-appliance/#howto-upgrade-appliance) for instructions on how to update your Anbox Cloud deployment to the 1.24.0 release.
--- a/reference/release-notes/1.24.0.md
+++ b/reference/release-notes/1.24.0.md
@@ -36,6 +36,7 @@ See the [Requirements](https://documentation.ubuntu.com/anbox-cloud/en/latest/re
 
 * This release includes Android 14 based AOSP and AAOS images.
 * We now support using the lock screen in Anbox Cloud([LP 1983426](https://bugs.launchpad.net/anbox-cloud/+bug/1983426)).
+* The out of band v2 feature in Anbox Cloud requires the Android app to be running to receive broadcasts. If the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle), the system places [context-registered broadcasts in a queue]((https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14)) and hence the app may not receive broadcasts immediately as it would when it is running.<!--AC-2852-->
 
 ### Charms
 
@@ -73,7 +74,6 @@ Following deprecation in earlier releases, these functionalities have been remov
 
 ## Known issues
 
-* Out of band support does not work as expected for Android 14 images because of a change in implicit intents feature in the upstream.<!--AC-2852-->
 * The following features can be enabled for AAOS images based on Android 14. However, this is not supported and may result in unexpected behaviour. This will be removed in the next patch release:<!--AC-2848-->
   - `enable_virtual_keyboard`
   - `enable_system_ui`
--- a/reference/release-notes/1.24.0.md
+++ b/reference/release-notes/1.24.0.md
@@ -36,7 +36,7 @@ See the [Requirements](https://documentation.ubuntu.com/anbox-cloud/en/latest/re
 
 * This release includes Android 14 based AOSP and AAOS images.
 * We now support using the lock screen in Anbox Cloud([LP 1983426](https://bugs.launchpad.net/anbox-cloud/+bug/1983426)).
-* The out of band v2 feature in Anbox Cloud requires the Android app to be running to receive broadcasts. If the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle), the system places [context-registered broadcasts in a queue]((https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14)) and hence the app may not receive broadcasts immediately as it would when it is running.<!--AC-2852-->
+* For the Android 14 image, the out of band v2 feature in Anbox Cloud requires the Android app to be running to receive broadcasts. If the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle), the system places [context-registered broadcasts in a queue]((https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14)) and hence the app may not receive broadcasts immediately as it would when it is running.<!--AC-2852-->
 
 ### Charms
 
@@ -70,7 +70,7 @@ Following deprecation in earlier releases, these functionalities have been remov
 
 ## Deprecations
 
-< List deprecated features. Provide an alternative, if any, for the users to use instead of the deprecated features >
+There are no new deprecations announced for 1.24.0. For the list of features or components that were deprecated before 1.24.0 and are planned to be removed in future releases, see {ref}`ref-deprecation-notes`.
 
 ## Known issues
 
@@ -81,6 +81,8 @@ Following deprecation in earlier releases, these functionalities have been remov
   - Custom Android ID
   - Custom system apps
   - Boot package support
+
+* When using the Android 14 image, displaying an activity from a boot package using the `foregroundActivity` option in the [Anbox Streaming SDK](https://github.com/canonical/anbox-streaming-sdk/tree/main) may not work. This issue occurs due to a native crash occurring in the Android container.
 
 ## Bug fixes
 
--- a/reference/release-notes/1.24.0.md
+++ b/reference/release-notes/1.24.0.md
@@ -27,6 +27,7 @@ See the [Requirements](https://documentation.ubuntu.com/anbox-cloud/en/latest/re
 * We have enhanced security with a minimum version of TLS 1.3 as a requirement for all services working with Anbox Cloud. TLS 1.3 is also enabled by default for the Anbox Stream Agent and Gateway.<!--AC-2728 and AC-2654-->
 
 ### Anbox Management Service (AMS)
+
 * We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
 * Anbox Cloud provides the image type (`android`, `aaos` or `generic`) through the following three ways:<!--AC-2757-->
   - `metadata.yaml` of the Anbox Cloud image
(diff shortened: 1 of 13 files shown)


reference/release-notes/release-notes.md
```diff
@@ -9,22 +9,23 @@ For instructions on how to update your Anbox Cloud deployment to later versions,
...
 
 The following target dates for upcoming releases are not final and could vary depending on various factors such as availability of Android security patches. The release notes link will be updated on the day of the release.
 
 | Target date | Version | Planned updates |
 |----|----|----|
-| November 14 2024 | Anbox Cloud 1.24.0 | * VHAL improvements<br/> * Android 14 images<br/> * Support for additional upscaling algorithms in the streaming SDK<br/> * Android security updates for November 2024<br/> * Reworked charms<br/> * Bug fixes |
 | December 11 2024 | Anbox Cloud 1.24.1 | * Android security updates for December 2024<br/> * Bug fixes |
 | January 15 2025 | Anbox Cloud 1.24.2 | * Android security updates for January 2025<br/> * Bug fixes |
+| February 12 2025 | Anbox Cloud 1.25.0 | * Integration with the Canonical Observability Stack<br/> * Backup and restore support for the Anbox Cloud dashboard charm<br/> * Support for Ubuntu 24.04 (Noble Numbat) for all charms<br/> * Terraform plan replacing the Anbox Cloud bundles<br/> * Improvements to the Anbox Management Service (AMS)<br/> * Upgrade to LXD 5.21<br/> * Android security updates for February 2025<br/> * Bug fixes |
```
keirthana, line 28: @morphis This is from Jira, please confirm if we are okay with highlighting listed features for 1.25.

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+## Removed functionality
+
+Following deprecation in earlier releases, these functionalities have been removed from Anbox Cloud with the 1.24.0 release:
+
+* Support for Ubuntu 20.04 (Focal Fossa)
+* Use of node controller to manage the port forwarding from an instance to expose a service
+
+## Deprecations
+
+< List deprecated features. Provide an alternative, if any, for the users to use instead of the deprecated features >
```
keirthana, line 67: Any deprecations this release?

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+
+* Android security updates for November 2024 (see [Android Security Bulletin - November 2024](https://source.android.com/docs/security/bulletin/2024-11-01) for more information).
+* The Android WebView has been updated to 130.0.6723.73.
+
+## Removed functionality
+
+Following deprecation in earlier releases, these functionalities have been removed from Anbox Cloud with the 1.24.0 release:
+
+* Support for Ubuntu 20.04 (Focal Fossa)
+* Use of node controller to manage the port forwarding from an instance to expose a service
```
keirthana, line 63: @adglkh We need to discuss whether to remove all information about node controller across documentation or just this specific scenario.

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Streaming improvements
+
+* Anbox Cloud now provides an option to allow the use of custom fragment shader-based upscaling algorithms for video streaming. This helps you to apply multiple shaders sequentially to perform multi-pass upscaling and replaces the default [AMD FidelityFX Super Resolution 1.0](https://gpuopen.com/fidelityfx-superresolution/) shader. See {ref}`sec-custom-fragment-shader` for more information.<!--AC-2541-->
+* The Anbox Streaming SDK now uses the [Snapdragon™ Game Super Resolution(SGSR)](https://github.com/SnapdragonStudios/snapdragon-gsr) algorithm, which integrates upscaling and sharpening in one single GPU shader pass, by default.<!--AC-2543-->
+
+### Other
+
+* Android security updates for November 2024 (see [Android Security Bulletin - November 2024](https://source.android.com/docs/security/bulletin/2024-11-01) for more information).
+* The Android WebView has been updated to 130.0.6723.73.
```
keirthana, line 56: This is based on https://github.com/canonical/anbox/pull/369. Please confirm.

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+
+See the [Requirements](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/requirements/) for details on general and deployment specific requirements to run Anbox Cloud.
+
+## New features & improvements
+
+### Updates to components
+
+* Anbox Cloud includes and uses the following latest versions for components it uses:
+  - LXC version 6.0.2<!--AC-2812-->
+  - Mesa driver version 24.2.6<!--AC-2800-->
```
keirthana, line 22: Need confirmation on Mesa driver version as Jira story just says latest version, I took 24.2.6 from the upstream but is that the one we are on?

ajanon: We're running 24.2.4 in our case, I just checked
```suggestion
  - Mesa driver version 24.2.4<!--AC-2800-->
```

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+  - VirGL renderer version 1.9.0.<!--AC-2800-->
+
+### Security
+
+* We have enhanced security with a minimum version of TLS 1.3 as a requirement for all services working with Anbox Cloud. TLS 1.3 is also enabled by default for the Anbox Stream Agent and Gateway.<!--AC-2728 and AC-2654-->
+
+### Anbox Management Service (AMS)
+* We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
+* Anbox Cloud provides the image type (`android`, `aaos` or `generic`) through the following three ways:<!--AC-2757-->
+  - `metadata.yaml` of the Anbox Cloud image
```
ajanon, line 32: Do users need to know this? We don't have to hide it, but it may not be too useful for them to know this anyway (as they probably cannot change it easily).

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+  - API response from the AMS when querying image information
```
ajanon, line 33: Should we add a link to the API docs here?

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+  - Mesa driver version 24.2.6<!--AC-2800-->
+  - VirGL renderer version 1.9.0.<!--AC-2800-->
+
+### Security
+
+* We have enhanced security with a minimum version of TLS 1.3 as a requirement for all services working with Anbox Cloud. TLS 1.3 is also enabled by default for the Anbox Stream Agent and Gateway.<!--AC-2728 and AC-2654-->
+
+### Anbox Management Service (AMS)
+* We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
+* Anbox Cloud provides the image type (`android`, `aaos` or `generic`) through the following three ways:<!--AC-2757-->
```
ajanon, line 31: I think `generic` is not something we officially support for users, and more of an internal thing for now.
@adglkh @morphis Wdyt?

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+  - `metadata.yaml` of the Anbox Cloud image
+  - API response from the AMS when querying image information
+  - Output of `amc image show` command
+
+  This allows the web client to adapt its interface according to the image type.
+* We have enabled support for multiple cluster management by AMS.<!--AC-2674-->
+
+### Android
+
+* This release includes Android 14 based AOSP images.
```
ajanon, line 41: And AAOS:
```suggestion
* This release includes Android 14 based AOSP and AAOS images.
```

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+* We now support using the lock screen in Anbox Cloud([LP 1983426](https://bugs.launchpad.net/anbox-cloud/+bug/1983426)).
+
+### Charms
+
+* For the NFS Operator charm for Anbox Cloud, `Cachefilesd` is now optional and is used only when `fsc` is provided as a mount option.<!--AC-2839-->
+
+### Streaming improvements
+
+* Anbox Cloud now provides an option to allow the use of custom fragment shader-based upscaling algorithms for video streaming. This helps you to apply multiple shaders sequentially to perform multi-pass upscaling and replaces the default [AMD FidelityFX Super Resolution 1.0](https://gpuopen.com/fidelityfx-superresolution/) shader. See {ref}`sec-custom-fragment-shader` for more information.<!--AC-2541-->
+* The Anbox Streaming SDK now uses the [Snapdragon™ Game Super Resolution(SGSR)](https://github.com/SnapdragonStudios/snapdragon-gsr) algorithm, which integrates upscaling and sharpening in one single GPU shader pass, by default.<!--AC-2543-->
```
ajanon, line 51: IIRC we don't enable it by default. Should we change 'uses' to something more generic, like 'provides'?
```suggestion
* The Anbox Streaming SDK now provides the [Snapdragon™ Game Super Resolution(SGSR)](https://github.com/SnapdragonStudios/snapdragon-gsr) algorithm, which integrates upscaling and sharpening in one single GPU shader pass, by default.<!--AC-2543-->
```
@adglkh

reference/release-notes/1.24.0.md
ajanon: This is accurate.

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+* Support for Ubuntu 20.04 (Focal Fossa)
+* Use of node controller to manage the port forwarding from an instance to expose a service
+
+## Deprecations
+
+< List deprecated features. Provide an alternative, if any, for the users to use instead of the deprecated features >
+
+## Known issues
+
+* Out of band support does not work as expected for Android 14 images because of a change in implicit intents feature in the upstream.<!--AC-2852-->
```
ajanon, line 71: @adglkh Would you mind sharing the workaround you have found?

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,96 @@
...
+* The following features are supported for AAOS images based on Android 14. This is not an intended support and will be removed in the next patch release:<!--AC-2848-->
```
ajanon, line 72: It's not really 'supported' then, right? ;)
Also, it's not as good as I wrote it in the Jira story tbh: most of these are not working as intended, it's more of a partial, buggy support.
How should we phrase this? How about something along those lines:
```suggestion
* The following features can be enabled for AAOS images based on Android 14. This is not supported and may result in unexpected behavior. This will be removed in the next patch release:<!--AC-2848-->
```
Wdyt?

ajanon (changes requested):
Thanks for taking care of the known issues!
Here are some comments from me.

reference/release-notes/1.24.0.md
adglkh: The argument above is not accurate. 

I think we can point people to the following page
https://developer.android.com/about/versions/14/behavior-changes-all#pending-broadcasts-queued
Specifically.
```
On Android 14, the system can [place context-registered broadcasts in a queue](https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14) while the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle).
```
This is why the app didn’t receive the broadcast, even though it was sent out by the broadcast-data-channels-status action. This also means that any app or service intended to receive the com.canonical.anbox.BROADCAST_DATA_CHANNELS_STATUS broadcast must remain in the running state, not in a **cached** state. This is why our test case fails: the app wasn’t running at system startup.

So I don't think it's a known issue but a caveat if people use out of band v2 feature in Android 14

reference/requirements.md
```diff
@@ -44,7 +44,6 @@ For external access to the Anbox Cloud Appliance, you must expose a couple of ne
 
 The Anbox Cloud Appliance supports the following Ubuntu versions:
 
-* Ubuntu 20.04 (Focal Fossa)
```
adglkh, line 47: Technically, our old application still supports running on 20.04 given that we will still roll out old appliance(epoch=0) in 1.24 release .

tutorial/installing-appliance.md
```diff
@@ -18,7 +18,7 @@ This tutorial guides you through the steps that are required to install and init
 Make sure you have the following prerequisites:
 
 * An Ubuntu SSO account. If you don't have one yet, [create it](https://login.ubuntu.com).
-* A virtual or bare metal machine running Ubuntu 20.04 or 22.04. See the detailed {ref}`ref-requirements`.
```
adglkh, line 21: See above

reference/release-notes/release-notes.md
```diff
@@ -45,18 +46,16 @@ To ensure you receive latest security updates and bug fixes, you should upgrade
...
-* Seamless ADB connections to remote Android instances
-* Dashboard improvements
-    - Merge of sessions and instances pages
-    - Ability to authorise ADB connections
-* Improved documentation delivery
-* Android 14 based vendor images
-* Reworked charms that install snaps from the snap store
+* VHAL improvements
+* Android 14 images
+* Support for generic application streaming
```
adglkh, line 55: Does this mean we're going to release a generic image in 1.24.0? if not, do we need to call ou in the release announcement?

reference/release-notes/1.24.0.md
adglkh: Sorry, I don't follow this. 
1. are we all agreeing to remove those features?
2. if so, are we going to do so in the next patch release?

reference/release-notes/1.24.0.md
adglkh: Because of [this](https://github.com/canonical/anbox-cloud-docs/pull/212/files#r1831507950)
We need to mention the deprecation of the node controller in this section.

reference/release-notes/1.24.0.md
adglkh: Yes, I think the following sounds better.

The Anbox Streaming SDK includes two additional examples that demonstrate shader-based upscaling algorithms integrated with the JS SDK :
- [Snapdragon™ Game Super Resolution(SGSR)](https://github.com/SnapdragonStudios/snapdragon-gsr)
- [Anime4K high-quality real-time anime](https://github.com/bloc97/Anime4K)

adglkh (changes requested):

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,105 @@
...
+### Anbox Management Service (AMS)
+
+* We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
+* Anbox Cloud provides the image type (`android`, `aaos` or `generic`) through the following three ways:<!--AC-2757-->
+  - `metadata.yaml` of the Anbox Cloud image
+  - API response from the AMS when querying image information
+  - Output of `amc image show` command
+
+  This allows the web client to adapt its interface according to the image type.
+* We have enabled support for multiple cluster management by AMS.<!--AC-2674-->
```
adglkh, line 38: This is not accurate.

The story is about enabling dashboard to communicate with multiple AMS instances.
However when I looked at the AC-2674, I didn't find a PR yet. So I am not quite sure if this feature is enabled or not in this cycle.
@jat-canonical  could you clarify on this as well?
Thanks

adglkh (changes requested):

reference/release-notes/1.24.0.md
keirthana: Okay so we don't have to list it here for 1.24.0 because it is already in the deprecation notices ( See https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/deprecation-notices/). I will adjust the removal target release to 1.25.0 then.

reference/release-notes/1.24.0.md
keirthana: It is already in the deprecation notice since 1.23.0: https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/deprecation-notices/. Since it is not yet ready for removal, I will keep this in place until 1.25.0

reference/release-notes/1.24.0.md
keirthana: Thanks for the clarification

reference/release-notes/1.24.0.md
keirthana: Not relevant as the section is rewritten

reference/release-notes/1.24.0.md
keirthana: I have removed generic from the image type now but for the generic application streaming, let's confirm with @morphis

reference/release-notes/1.24.0.md
keirthana: As far as I understand, these features should not be available for enabling but they are. We intend to remove these in the next patch release.

reference/release-notes/release-notes.md
keirthana: @morphis Can you confirm if generic application streaming is dropped from 1.24.0?

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Anbox Management Service (AMS)
+
+* We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
+* Anbox Cloud provides the image type (`android`, `aaos` or `generic`) through the following three ways:<!--AC-2757-->
+  - `metadata.yaml` of the Anbox Cloud image
+  - API response from the AMS when querying image information
+  - Output of `amc image show` command
+
+  This allows the web client to adapt its interface according to the image type.
```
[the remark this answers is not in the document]
keirthana: Not relevant as the feature is rewritten without these details

reference/requirements.md
keirthana: Hmm, but according to our [deprecation notices](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/deprecation-notices/#ubuntu-20-04-focal), we are supposed to drop 20.04 support from 1.24.0. 

Yes the appliance epoch=0 will be around until 1.25.0 but I think we can start dropping official support for focal. WDYT @morphis ?

reference/release-notes/1.24.0.md
keirthana: @adglkh We have two options:

1. Drop the known issue as it is actually a caveat like you mentioned.
2. Include the change under Android section of features and improvements (yes, it is not necessarily a feature or improvement) since something has changed in the upstream behaviour and has caused a ripple effect in Anbox as well. We can list something like:
    > The out of band feature in Anbox Cloud works differently with Android 14 because of changes in upstream. On Android 14, the system can [place context-registered broadcasts in a queue](https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14) while the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle). This means that an app needs to be running for it to receive the broadcast.

WDYT?

adglkh: > @adglkh We have two options:
> 
> 1. Drop the known issue as it is actually a caveat like you mentioned.
> 2. Include the change under Android section of features and improvements (yes, it is not necessarily a feature or improvement) since something has changed in the upstream behaviour and has caused a ripple effect in Anbox as well. We can list something like:
>    > The out of band feature in Anbox Cloud works differently with Android 14 because of changes in upstream. On Android 14, the system can [place context-registered broadcasts in a queue](https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14) while the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle). This means that an app needs to be running for it to receive the broadcast.
> 
> WDYT?

A: That sounds good to me. Let me reword this a bit
```
The out of band v2 feature in Anbox Cloud requires the Android app to be running to receive the broadcast If Android 14 image is in use due to the changed behavior that the system [place context-registered broadcasts in a queue]((https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14)) while the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle).
```
Please adjust this when needed.

reference/release-notes/release-notes.md
adglkh: Hmm, we’ve never formally released the generic image, so this isn't about dropping it as of version 1.24.0.

reference/release-notes/release-notes.md
adglkh: @morphis @jat-canonical correct me if I'm mistaken, but as discussed in the sprint, the Terraform plan work has been moved from the February 2025 timeline to May 2025 (and we agreed that the TF work should be ready by early April).

keirthana: Yes @adglkh, Terraform shouldn't be in there. I am waiting for other feedback on this to push the changes together.

morphis: We can keep it in here but label it as an early "alpha" release. The plan might be maintained independently of our regular releases.

reference/requirements.md
morphis: @adglkh technically a lot is possible :-) However from a support/maintenance perspective all support for focal is gone by now.

reference/release-notes/release-notes.md
morphis: @keirthana generic images have never been released and will not any time soon. So lets not mention them at all.

tutorial/installing-appliance.md
keirthana: As per Simon's reply above, we will keep the removal of focal support in explicit mentions.

ajanon (approved):
LGTM

morphis (changes requested):

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,101 @@
...
+* Support for Ubuntu 20.04 (Focal Fossa)
+
+## Deprecations
+
+< List deprecated features. Provide an alternative, if any, for the users to use instead of the deprecated features >
+
+## Known issues
+
+* Out of band support does not work as expected for Android 14 images because of a change in implicit intents feature in the upstream.<!--AC-2852-->
+* The following features can be enabled for AAOS images based on Android 14. However, this is not supported and may result in unexpected behaviour. This will be removed in the next patch release:<!--AC-2848-->
```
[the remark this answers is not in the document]
ajanon: As in, saying that we share the same source tree this time around? Is it useful for users to know?

reference/requirements.md
adglkh: Okay, let's drop that

tutorial/installing-appliance.md
adglkh: Yes, that LGTM.
Thanks

reference/release-notes/release-notes.md
```diff
@@ -45,18 +46,16 @@ To ensure you receive latest security updates and bug fixes, you should upgrade
...
+* Support for additional upscaling algorithms in the streaming SDK
+* Support for managing multiple subclusters
+* Reworked charms
```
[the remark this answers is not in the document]
ajanon: * Support for additional upscaling algorithms in the streaming SDK

Isn't this in the SDK, though? We have the snapdragon one and the AMD one, right?

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,101 @@
...
+
+* We have increased the maximum number of queued operations in AMS to allow better performance.<!--AC-2831-->
+* AMS supports querying the image variant (`android` or `aaos`) for Anbox Cloud images.<!--AC-2757-->
+* The web client, for example, the Anbox Cloud dashboard can communicate with multiple AMS instances.<!--AC-2674-->
+
+### Android
+
+* This release includes Android 14 based AOSP and AAOS images.
+* We now support using the lock screen in Anbox Cloud([LP 1983426](https://bugs.launchpad.net/anbox-cloud/+bug/1983426)).
+* The out of band v2 feature in Anbox Cloud requires the Android app to be running to receive broadcasts. If the app is in the [cached state](https://developer.android.com/guide/components/activities/process-lifecycle), the system places [context-registered broadcasts in a queue]((https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14)) and hence the app may not receive broadcasts immediately as it would when it is running.<!--AC-2852-->
```
adglkh, line 39: The out of band v2 feature in Anbox Cloud requires the Android app to be running to receive broadcasts when using the Android 14 LXD image. In short, this is Android 14 specific.
That implies if people are using Android 12 or Android 13 LXD images, apps can stay cache state and the out of band v2 feature worked fine.

reference/release-notes/release-notes.md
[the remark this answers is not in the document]
adglkh: I guess Simon specifically meant the `Reworked charms` item.

adglkh (commented):
One last comment, otherwise LGTM.
Thanks

reference/release-notes/1.24.0.md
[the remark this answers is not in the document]
keirthana: It is a technical implementation detail, no? I don't think this will be useful.

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Removed functionality
+
+Following deprecation in earlier releases, these functionalities have been removed from Anbox Cloud with the 1.24.0 release:
+
+* Support for Ubuntu 20.04 (Focal Fossa)
+
+## Deprecations
+
+< List deprecated features. Provide an alternative, if any, for the users to use instead of the deprecated features >
```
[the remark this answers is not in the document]
keirthana: I wouldn't do it that way because it switches the context of the release notes which is to focus on the current release. So any deprecations we are announcing in 1.24.0 will go here. 

I have added a statement that, in addition to calling out no deprecations for 1.24.0, will remind users to look at our deprecation notices to prepare for next releases.

reference/release-notes/1.24.0.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Known issues
+
+* The following features can be enabled for AAOS images based on Android 14. However, this is not supported and may result in unexpected behaviour. This will be removed in the next patch release:<!--AC-2848-->
+  - `enable_virtual_keyboard`
+  - `enable_system_ui`
+  - `enable_anbox_ime`
+  - Custom Android ID
+  - Custom system apps
+  - Boot package support
```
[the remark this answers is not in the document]
keirthana: Done, thanks!

adglkh (approved):
LGTM, thanks!

morphis (approved):
LGTM

PR #215, opened 2024-11-12
fix(howto): update troubleshooting to newer appliance commands

morphis:
This was forgotten when the appliance documentation was changed.


keirthana (approved):
LGTM

PR #218, opened 2024-11-14
fix(troubleshoot): use `juju-crashdump` over `juju crash-dump` command

adglkh:
Juju 3 drops support for executing the external plug, which causes
the juju crash-dump command to be incomplete in Juju 3 and results
in the following error:
```
$ juju crashdump
ERROR juju: "crashdump" is not a juju command. See "juju --help".
``

Therefore, use the juju-crashdump command in our documentation to
ensure users can collect logs regardless of the Juju version.

Another issue with the suggested command is that we need to include
the `debug-layer` addon followed by the -a parameter, which denotes
the --addon option.

### DIFF
--- a/howto/troubleshoot/landing.md
+++ b/howto/troubleshoot/landing.md
@@ -36,7 +36,7 @@ A Juju crash dump may include the following debugging information:
 
 Use the following command to generate a crash dump:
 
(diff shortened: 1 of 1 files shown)


ajanon (approved):
LGTM

morphis (approved):
LGTM

PR #219, opened 2024-11-14
Improve guide for TLS configuration of the appliance

morphis:
I took a test run on the current how-to and had to adjust a few things to make it work after recent changes we did to the appliance.  We miss one bug fix for AMS to make it fully work which will come with 1.24.1 but will be soon in 1.24/beta. Regardless of this we should update the documentation to reflect how TLS should be setup.

### DIFF
--- a/howto/anbox/tls-for-appliance.md
+++ b/howto/anbox/tls-for-appliance.md
@@ -11,7 +11,7 @@ Before you start, make sure the following requirements are met:
 
 - The Anbox Cloud Appliance is installed and initialised. See {ref}`howto-install-appliance-aws` and {ref}`sec-initialise-appliance` for instructions.
 
-- If you want to use the AWS certificate manager, you should have a domain name for use with the Anbox Cloud Appliance registered, either through Amazon Route 53 (see [Register a new domain](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/domain-register.html) in the Amazon Route 53 documentation) or  through another domain provider.
+- A DNS name you can use to point to the machine hosting the Anbox Cloud Appliance
 
 ## Manage the certificate manually
 
@@ -19,7 +19,7 @@ To generate and install a certificate yourself, complete the following steps:
 
 ### Add a DNS record
 
-Setting up DNS redirection depends on your DNS provider. Refer to the documentation of your provider to create a DNS record pointing to the IP/DNS of the AWS instance where the Anbox Cloud Appliance is running.
+Setting up DNS redirection depends on your DNS provider. Refer to the documentation of your provider to create a DNS record pointing to the IP/DNS of the machine where the Anbox Cloud Appliance is running.
 
 
 (ref-appliance-tls-location)=
@@ -38,6 +38,8 @@ There are many ways to create a valid SSL certificate. One way is to use [Let's 
 First, connect and SSH into your appliance instance, and install the `certbot` snap:
 
     sudo snap install --classic certbot
+
+Using certbot requires that TCP port 80 is not in use as certbot will use it to verify your ownership of the requested DNS name. The Anbox Cloud Appliance only occupies TCP port 443. Please ensure that nothing is listening on the machine on TCP port 80 before you continue.
 
 Then run the following command to generate your certificate:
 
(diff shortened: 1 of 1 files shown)


adglkh (approved):
I agree that the section on using AWS Certificate Manager doesn't contain any Anbox Cloud specific content. However, it still provides value for people who want to use AWS Certificate Manager to create certificates and set things up for an appliance running on an EC2 instance (particularly for the `allows only HTTPS traffic` setting and the default outbound rule that allows all traffic).

Otherwise LGTM

howto/anbox/tls-for-appliance.md
```diff
@@ -21,43 +21,15 @@ To generate and install a certificate yourself, complete the following steps:
...
-This option is experimental. It will be removed in a future release when a better replacement exists.
-```
 
 (ref-appliance-tls-location)=
 ### Configure the location
 
-Configure the location for the appliance using the created DNS name. For the following configuration files need to be updated:
-
-#### `/var/snap/anbox-cloud-appliance/common/gateway/config.yaml`
+Configure the location for the appliance using the created DNS name by running the following command
```
keirthana, line 28:
```suggestion
Configure the location for the appliance using the created DNS name by running the following command:
```

howto/anbox/tls-for-appliance.md
```diff
@@ -21,43 +21,15 @@ To generate and install a certificate yourself, complete the following steps:
...
-```yaml
-log-level: error
-listen-address: 192.168.11.195
-location: your.dns.name
-...
-tls:
-  # Drop or comment the following line
-  # ca: /var/snap/anbox-cloud-appliance/common/tls/ca.crt
-```
+The change will be automatically applied and causes all services components of the appliance to restart. If you want to defer the restart to a later point, you can use the `--no-restart` option.
```
keirthana, line 32:
```suggestion
The change will be automatically applied and will cause all services components of the appliance to restart. If you want to defer the restart to a later point, you can use the `--no-restart` option.
```

howto/anbox/tls-for-appliance.md
```diff
@@ -93,167 +61,26 @@ Copy the generated certificate to the `/var/snap/anbox-cloud-appliance/common/da
...
-   ```bash
-   cat <<EOF | sudo tee /etc/letsencrypt/renewal-hooks/pre/001-stop-appliance.sh
-   #!/bin/bash
-   sudo snap stop anbox-cloud-appliance.daemon
-   EOF
-   sudo chmod +x /etc/letsencrypt/renewal-hooks/pre/001-stop-appliance.sh
-   ```
-
-2. Install the certificate right after it has been renewed and start the appliance service through the `post-hook`:
+The `certbot` snap packages installed on your machine already set up a systemd timer that will automatically renew your certificates before they expire. However, in order to get the certificate renewed successfully for the appliance, you can create `post-hook` hook for certbot which will automatically reconfigure it:
```
keirthana, line 72: Did you mean a post-start hook?

```suggestion
The `certbot` snap packages installed on your machine would have already set up a systemd timer that automatically renews your certificates before they expire. However, to get the certificate renewed successfully for the appliance, you can create `post-start` hook for `certbot` which will automatically reconfigure it:
```

howto/anbox/tls-for-appliance.md
```diff
@@ -93,167 +61,26 @@ Copy the generated certificate to the `/var/snap/anbox-cloud-appliance/common/da
...
+   sudo snap restart anbox-cloud-appliance.daemon
    EOF
    sudo chmod +x /etc/letsencrypt/renewal-hooks/post/001-start-appliance.sh
    ```
 
-In this way, the SSL certificate auto-renewal is in place.
-
 ```{note}
-The appliance will face a short downtime during the renewal of the SSL certificate but will come back online once the process is completed.
+The appliance will be restarted at the end of the renewal of the SSL certificate to let the reverse proxy reload the certificate.
```
keirthana, line 85:
```suggestion
The appliance will be restarted when the renewal of the SSL certificate is complete, to let the reverse proxy reload the certificate.
```

keirthana (changes requested):
Some minor suggestions

adglkh:
> > I agree that the section on using AWS Certificate Manager doesn't contain any Anbox Cloud specific content. However, it still provides value for people who want to use AWS Certificate Manager to create certificates and set things up for an appliance running on an EC2 instance (particularly for the `allows only HTTPS traffic` setting and the default outbound rule that allows all traffic).
> 
> It may but the bigger problem is that it's hard for us to verify it doesn't break and still works. We seem to suffer a bit from content at the moment which is untested and turns out to be actually not helpful. Thus we need to be careful IMHO with what content we provide and we can actually keep working. The AWS UI has for example recently changed and I don't know if it is much longer the same way of enabling the TLS setup the document was describing. Especially as this is an edge case of non-Anbox Cloud functionality we should not keep it but much rather refer to external documentation (e.g. AWS) if needed. Wdyt?

A: Okay, let's drop the `using AWS Certificate Manager` section now and revisit it if necessary.

ajanon (approved):
LGTM

keirthana (approved):

PR #220, opened 2024-11-14
Fetch gateway swagger API document automatically

morphis:
This extends the existing process we have for the AMS API to also automatically fetch the gateway API document from a running appliance following the current release of Anbox Cloud.


adglkh (approved):
LGTM

PR #221, opened 2024-11-15
fix(ref): put links into code blocks to avoid linkcheck

morphis:
The addresses used in the examples should not be checked by the link checker as they are not accessible and cause the check to timeout.


keirthana:
@morphis These files are autogenerated and hence this needs to be fixed at the source. Else they will get overwritten the next time the command reference is generated. See https://github.com/canonical/anbox-cloud-docs/blob/main/.github/workflows/generate-command-reference.yml

keirthana (changes requested):
This should not be updated here but rather in the appliance source code since this is generated by running the `sudo anbox-cloud-appliance generate documentation -f markdown` command

PR #224, opened 2024-11-19
Fix documentation around --enable-graphics flag

### LINKED ISSUE #216: docs: it makes confusing
*Please describe the question or issue you're facing with "GPUs and instances - Anbox Cloud documentation".*

It has an inconsistent description below screenshot.

chrome_screenshot_2024  11  13  오전 9시 38분 30초 GMT+09_00

---
*Reported from: https://documentation.ubuntu.com/anbox-cloud/en/latest/explanation/gpus-instances/*

keirthana:
Fixes #216


explanation/gpus-instances.md
```diff
@@ -1,38 +1,43 @@
...
 ```
 
-Anbox Cloud automatically detects GPU devices during the deployment and configures the cluster to use them. If no GPU is available, Anbox Cloud automatically falls back to the `null` platform that does not perform any rendering. See {ref}`sec-null-platform` for the `null` platform configuration. 
+Anbox Cloud automatically adjusts its behaviour depending on the availability of a GPU and the presence of the `--enable-graphics` flag when you launch an instance:
 
-However, you can enable software rendering and video encoding by launching your application with the `--enable-graphics` flag. This makes it possible to run entirely without a GPU and still use rendering.
+    amc launch --enable-graphics my-application
 
-## Required GPU slots
+The use of `--enable-graphics` flag when you launch an instance decides which platform (WebRTC or `null`) is used for rendering.
```
adglkh, line 14: which platform... ->
->
which [platform](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/supported-rendering-resources/#supported-platforms) ...

explanation/gpus-instances.md
```diff
@@ -1,38 +1,43 @@
...
+- Check the list of supported GPUs ({ref}`sec-supported-gpus`) to see if Anbox Cloud includes a driver for your GPU device. If a GPU driver is available inside the instance, there are no further differences in how to use it in comparison to a regular environment. If no GPU driver is available, you must provide it through an addon.
+- Launch your instance with the `--enable-graphics` flag.
 
-If you want to let an application use the GPU (even if you are not interested in streaming the visual output), launch it with the `--enable-graphics` flag. With this flag, the command will launch the instance using the `webrtc` platform, which will automatically detect the underlying GPU and make use of it.
+(sec-sw-rendering-video-encoding)=
+## Without a GPU
 
-    amc launch --enable-graphics my-application
+Anbox Cloud provides a way to run entirely without GPU by using software rendering. You can enable software rendering and video encoding by launching your application with the `--enable-graphics` flag. When there is no GPU and this flag is enabled, Anbox Cloud falls back to using the `null` platform.
```
adglkh, line 33: Anbox Cloud provides a way to run entirely without GPU by using software rendering. You can ....
->
Anbox Cloud supports deployment in environments without GPUs by utilizing software or headless rendering. In such environments, when no GPU is available, you can ...

For the context, here we may need to focus on **a deployment without GPU**. Because in a deployment with GPU, an application with instance type `a2.3` or no gpu-slots is allocated, people can still enable software rendering by using
`amc launch --enable-graphis <app>` command.

adglkh (changes requested):

adglkh (commented):
While reading this page, another question came to mind: with the recently added CLI flag `--enable-streaming` to `amc launch` comman, the `webrtc` platform will be used in such cases. Since the `webrtc` platform automatically detects the GPU and adjusts rendering or video encoding capabilities (software vs hardware) accordingly, should we mention this on this page as well?

explanation/gpus-instances.md
keirthana: > For the context, here we may need to focus on a deployment without GPU. Because in a deployment with GPU, an application with instance type a2.3 or no gpu-slots is allocated, people can still enable software rendering by using
`amc launch --enable-graphis <app> command.`

Yes the part about with the GPU is covered in the earlier section.

keirthana:
> While reading this page, another question came to mind: with the recently added CLI flag `--enable-streaming` to `amc launch` comman, the `webrtc` platform will be used in such cases. Since the `webrtc` platform automatically detects the GPU and adjusts rendering or video encoding capabilities (software vs hardware) accordingly, should we mention this on this page as well?

I have added a note because it is worth informing the user that there is another flag (`--enable-streaming`) that could influence the rendering platform used.

ajanon (approved):
LGTM

adglkh (approved):
LGTM

explanation/gpus-instances.md
```diff
@@ -1,38 +1,47 @@
...
-Anbox Cloud automatically detects GPU devices during the deployment and configures the cluster to use them. If no GPU is available, Anbox Cloud automatically falls back to the `null` platform that does not perform any rendering. See {ref}`sec-null-platform` for the `null` platform configuration. 
+Anbox Cloud automatically adjusts its behaviour depending on the availability of a GPU and the presence of the `--enable-graphics` flag when you launch an instance:
 
-However, you can enable software rendering and video encoding by launching your application with the `--enable-graphics` flag. This makes it possible to run entirely without a GPU and still use rendering.
-
-## Required GPU slots
+    amc launch --enable-graphics my-application
 
-GPUs have limited capacity that can be shared amongst multiple instances. To fine-tune how many instances can run on a given node, configure the number of available GPU slots on the node.
+The use of `--enable-graphics` flag when you launch an instance decides which [platform](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/supported-rendering-resources/#supported-platforms)(WebRTC or `null`) is used for rendering.
```
morphis, line 14: Should we keep the explanation of the platform which is used in here? It's not really relevant to the user, right? We should much rather explain it from the perspective of what the outcome of specifying the flag is. With adding it we enable pixel rendering whereas without we run in a non-pixel rendering mode which has less CPU consumption but never produces any visual output

morphis (changes requested):
Overall reads much better but we should still get rid of the platform usage as IMHO it might add more confusion/involves another advanced topic the user might not care about to know at this point. Wdyt?

explanation/gpus-instances.md
adglkh: This is a bit confusing for me as well, because the entire chapter is going to explain the deployment with GPU equipped and influence the underlying instance with GPU support or not with `gpu-slots`. We could mention the `graphics-headless` or visual output when the --enable-graphics flag is used or not, but mentioning the CPU here seems unnecessary.

adglkh:
> get rid of the platform usage as IMHO it might add more confusion/involves another advanced topic the user might not care about to know at this point. Wdyt?

A: Yeah, sort of. However, if we want to remove the platform context here, we should avoid mentioning software and hardware rendering or video encoding (which are the focus of the chapter about using the GPU in instances). Instead, we should only refer to the visual output, which could be influenced by the `--enable-graphics` flag. In other word, the `--enable-graphics` flag doesn't directly determine whether the GPU is used or not; it only decides whether there is visual output from the instance.

morphis:
> > get rid of the platform usage as IMHO it might add more confusion/involves another advanced topic the user might not care about to know at this point. Wdyt?
> 
> A: Yeah, sort of. However, if we want to remove the platform context here, we should avoid mentioning software and hardware rendering or video encoding (which are the focus of the chapter about using the GPU in instances). Instead, we should only refer to the visual output, which could be influenced by the `--enable-graphics` flag. In other word, the `--enable-graphics` flag doesn't directly determine whether the GPU is used or not; it only decides whether there is visual output from the instance.

We are in a page titled "GPUs and instances" so why shouldn't we talk about hardware rendering and video encoding? Both are essential functionalities we use from a GPU.

We need to explain the --enable-graphics switch as even if you assign a GPU to the instance you won't get it being used by Anbox unless you have it set.

keirthana:
@morphis @adglkh I have rearranged the entire topic to focus on graphical rendering based on the `--enable-graphics` flag. Please have a look.

I intentionally removed the information about `--enable-streaming` flag that we had in the earlier round, as it spoke about the rendering platform rather than the purpose of use.

explanation/landing.md
```diff
@@ -37,7 +37,7 @@ Understand deployment configurations to plan your deployment.
 |--|--|
 | {ref}`exp-production-planning` | Aspects to consider for best results when you are planning a production deployment |
 | {ref}`exp-capacity-planning` | How to estimate the required capacity for your Anbox Cloud deployment |
-| {ref}`exp-gpus-instances` | How to manage GPUs and make them available to instances |
+| {ref}`exp-rendering-graphics` | How to manage GPUs and make them available to instances |
```
morphis, line 40: The title here needs to be updated 

```suggestion
| {ref}`exp-rendering-graphics` | Rendering graphical output |
```

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
+(exp-rendering-graphics)=
+# Rendering graphical output
+
+Anbox Cloud can adapt to your rendering needs by adjusting its behaviour depending on the availability of a GPU and the presence of the `--enable-graphics` flag when you launch an instance.
+
+    amc launch --enable-graphics my-application
+
+The use of `--enable-graphics` flag when you launch an instance decides whether Anbox Cloud renders a visual output.
```
morphis, line 8:
```suggestion
The use of the `--enable-graphics` flag when you launch an instance decides whether Anbox Cloud renders a visual output.
```

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
+(exp-rendering-graphics)=
+# Rendering graphical output
```
morphis, line 2: We don't use "graphical output" but "visual output" only. Is that planned?

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+
+Anbox Cloud can adapt to your rendering needs by adjusting its behaviour depending on the availability of a GPU and the presence of the `--enable-graphics` flag when you launch an instance.
+
+    amc launch --enable-graphics my-application
+
+The use of `--enable-graphics` flag when you launch an instance decides whether Anbox Cloud renders a visual output.
+
+When this flag is set, if there is a GPU available, Anbox Cloud renders visual output using the GPU. If there is no GPU available, it uses software rendering to render the visual output.
+
+However, if you do not use the `--enable-graphics` flag when launching an instance, Anbox Cloud does not render a visual output.
```
morphis, line 12:
```suggestion
However, if you do not use the `--enable-graphics` flag when launching an instance, Anbox Cloud does not provide any visual output.
```

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+
+Now your environment may or may not include a GPU. If Anbox Cloud detects a GPU device during deployment, it configures the cluster to use the GPU. AMS configures each LXD instance to pass through a GPU device from the host. Currently, all GPUs that are available to a machine are passed to every instance that owns a GPU slot. For NVIDIA GPUs, LXD uses the [NVIDIA container runtime](https://github.com/NVIDIA/nvidia-container-runtime) to make the GPU driver of the host available to the instance.
```
morphis, line 14:
```suggestion
Now your environment may or may not include a GPU. If Anbox Cloud detects a GPU device during deployment, it adjusts itself to use the GPU. AMS configures each LXD instance to pass through a GPU device from the host. Currently, all GPUs that are available to a machine are passed to every instance that owns a GPU slot. For NVIDIA GPUs, LXD uses the [NVIDIA container runtime](https://github.com/NVIDIA/nvidia-container-runtime) to make the GPU driver of the host available to the instance.
```

We don't always have a cluster so lets be more generic.

Furthermore `Currently, all GPUs that are available to a machine are passed to every instance that owns a GPU slot.` isn't true. We only pass a single GPU but the entire GPU will be available to the instance. There is no slicing support available.

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+
+```{important}
+Anbox Cloud currently does not support GPU provisioning for virtual machines. This feature is planned for a future release.
```
morphis, line 17: This can be read in different ways. If the virtual machine has a GPU attached and we run Anbox Cloud on top of it, we can use it. I think what you mean is that if we run a VM instance within AMS with an Anbox image, right?

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+To use container instances with a GPU,
+
+- Configure the number of available GPU slots on the node. This decides how many instances can run on a given node because GPUs have limited sharing capacity amongst multiple instances. See {ref}`sec-gpu-slots` for detailed information.
+- Check the list of supported GPUs ({ref}`sec-supported-gpus`) to see if Anbox Cloud includes a driver for your GPU device. If a GPU driver is available inside the instance, there are no further differences in how to use it in comparison to a regular environment. If no GPU driver is available, you must provide it through an {ref}`addon <exp-addons>`.
```
morphis, line 23: We don't include drivers in Anbox Cloud. Maybe

```suggestion
- Check the list of supported GPUs ({ref}`sec-supported-gpus`) to see if Anbox Cloud has support for a driver for your GPU device. If a GPU driver is available inside the instance, there are no further differences in how to use it in comparison to a regular environment. If no GPU driver is available, you must provide it through an {ref}`addon <exp-addons>`.
```

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+- Launch your instance with the `--enable-graphics` flag.
+
+If your environment does not have a GPU, Anbox Cloud uses software or headless rendering.
```
morphis, line 26: What do you mean by "headless rendering"? Headless means there is no output attached but it is not something which can replace "software rendering".

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+
+To use instances without a GPU and still render visual output,
+
+- Change the {ref}`sec-application-manifest-resources` of the application to not require a GPU. This tells Anbox Cloud that no GPU is available.
```
morphis, line 30:
```suggestion
- Change the {ref}`sec-application-manifest-resources` of the application to not require a GPU. This tells Anbox Cloud that no GPU should be used.
```

explanation/rendering-graphics.md
```diff
@@ -0,0 +1,38 @@
...
+- Launch your instance with the `--enable-graphics` flag. The presence of this flag when there is no GPU tells Anbox Cloud to use software rendering.
+
+Since software rendering and video encoding utilise the CPU, you won't be able to run as many instances on a system when compared to running instances with a GPU.
```
morphis, line 33: I would drop this as it depends on performance characteristics we don't have under control.

morphis (changes requested):

explanation/rendering-graphics.md
keirthana: Ok, so they are not interchangeable. Then, let's talk only about software rendering becauses that's what we use consistenly across our docs.

explanation/rendering-graphics.md
keirthana: Yes, the instances that are of type VM and not container - clarified the same.

morphis (approved):
LGTM!

morphis:
@adglkh do you want to take another look?

adglkh (approved):
LGTM

PR #225, opened 2024-11-20
Disable linkcheck until CLI text is fixed

keirthana:
Disabling the link check as it fails on invalid links often and also the CLI help text has links that trigger failures.
There are plans to fix the CLI help text shortly post which this change will be reverted.


ajanon (approved):
LGTM, do we have a story somehwere to remember to put it back on later on?

keirthana:
> LGTM, do we have a story somehwere to remember to put it back on later on?

https://warthogs.atlassian.net/browse/AC-2973 covers that and is scheduled for pulse 26. Hopefully by then, I should have addressed all issues.

adglkh (approved):
LGTM

PR #232, opened 2024-11-26
Add security policy to documentation

keirthana:
Since we have some private source code, this PR adds the policy to the documentation in addition to the `SECURITY.md` file to the public repositories.

### DIFF
--- a/reference/security-policy.md
+++ b/reference/security-policy.md
@@ -21,5 +21,10 @@ If you discover a security vulnerability with Anbox Cloud, report it using the f
 
 The [Ubuntu Security disclosure and embargo policy](https://ubuntu.com/security/disclosure-policy) contains more information about what you can expect when you contact us and what we expect from you.
 
-The Anbox Cloud team will be notified of the issue and review the vulnerability. We may reach out to you for further information or clarification if needed. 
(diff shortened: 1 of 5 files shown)


reference/security-policy.md
```diff
@@ -0,0 +1,25 @@
+(ref-security-policy)=
+# Security policy
+
+Learn about our [release and support policy](https://documentation.ubuntu.com/anbox-cloud/en/latest/reference/release-notes/release-notes/#release-and-support-policy) for the nature of our releases and versions.
```
morphis, line 4: Should we add in the linked document what we have defined [here](https://docs.google.com/document/d/1IkAanYou9y5sUIaLZC8JrIVTmpcBzi3JyE_bWT73dUo/edit?tab=t.0#heading=h.jf2awfdyd6r1) to let users know what we're going to fix and what not?

keirthana: I have now included details about scoring and our response process to the same document as this is the right place to talk about that.

See [line #24](https://github.com/canonical/anbox-cloud-docs/pull/232/files#diff-ae73e90592369962b66aa1444c3783ce5f560913042994c4d49ed1ad948c82a5R24)

PR #235, opened 2024-11-28
Clean up CLI help text created manually

keirthana:
This PR removes content that was created manually before we automated command reference topics and also fixes the pointer of the AMS configuration attributes content to the automatically created one.


jat-canonical (approved):
Overall LGTM. One small question about the amc command refs, shouldn't we do the same i.e. delete the `ams.amc` ones as well and put them into gitignore as they are also something that the workflow auto generates ?

keirthana:
> Overall LGTM. One small question about the amc command refs, shouldn't we do the same i.e. delete the `ams.amc` ones as well and put them into gitignore as they are also something that the workflow auto generates ?

So the way the command refs are generated are slightly different. They get generated every time there is an update/change and get updated in the GH repo through a PR (not during build time). However, the AMS configuration attributes files is generated during the documentation build time. So in order to avoid the source and published content from being out of sync with each other, removing this file in the repository is important.

However, that's not the case with the command references where the source is updated and then just published when RTD builds the docs.

PR #241, opened 2024-12-10
docs(tutorial): add warning to reboot system before running prepare-node-script

jat-canonical:
This is required as the kernel modules and drivers installed during the node setup should be installed in the correct and updated kernel on the machine.


keirthana (approved):
LGTM

ajanon (approved):
LGTM

PR #242, opened 2024-12-10
Clarify wording around upgrade

keirthana:
This PR clarifies the subtlety of upgrade vs update. We should use upgrade for when we mean moving to a different (later) version. Upgrading will result in a different version (be it patch or minor release) of the product.

Update should be used when we change something, but the resulting product version stays the same. For example: "Update the configuration in ..."

When we are referring to another product's upgrade mechanism, we should use their terminology but clarify that it is not Anbox Cloud's but rather the other product's mechanism we are referring to. For example: Controlling snap updates.


morphis (approved):
LGTM, great!

ajanon (approved):
LGTM

jat-canonical (approved):
LGTM

PR #243, opened 2024-12-10
Update regular to charmed deployment

keirthana:
This PR changes (almost) all the mentions of "regular" or "full" Anbox Cloud installation/deployment. We will be using the terminology "Charmed deployment" vs "appliance" to differentiate between our two variants.

I did a general sweep of the mentions and fixed them. If there are any left in specific context, we can fix them as we update the documentation.


morphis (approved):
This is fantastic! LGTM

PR #254, opened 2024-12-23
feat(streamer): add support for NVIDIA AV1 video encoder options in WebRTC streamer

adglkh:
Anbox Cloud has supported the AV1 video encoder for some time. The relevant
options for NVIDIA video encoders apply not only to H.264 but also to AV1.

This change introduces support for NVIDIA AV1 video encoder options in the
WebRTC streamer.


keirthana (approved):
LGTM, thanks!

PR #258, opened 2025-01-08
Enable link check and fix outdated links

keirthana:
The links from the CLI help text are fixed now. So the link check can be enabled now.
This PR enables the link check target and also fixes some outdated links in the documentation.

Jira ticket number: AC-2973


ajanon (approved):
LGTM

PR #259, opened 2025-01-09
update create and delete applications details

anusha-c18:
Add details for creating and deleting applications and creating instances.

### DIFF
--- a/howto/application/delete-application.md
+++ b/howto/application/delete-application.md
@@ -1,7 +1,9 @@
 (howto-delete-application)=
 # How to delete an application
 
-When an application is no longer needed, it can be fully removed from Anbox Cloud. Removing an application will cause all of its versions to be removed, including all of its currently active instances. Be extra careful as this might affect your users if any are still using instances of the application you want to remove.
+When an application is no longer needed, it can be fully removed from Anbox Cloud. However, be mindful of the consequences that come with removing an application such as:
+- Removal of associated base instances. Note that the base instance could first enter an error state before being cleaned up.
+- Removal of all the versions of the deleted application.
 
 Once you're sure you want to remove the application, you can delete it via the dashboard or the CLI.
 
--- a/howto/application/create-application.md
+++ b/howto/application/create-application.md
@@ -123,7 +123,13 @@ resources:
 
 ```{group-tab} Dashboard
 
-Click *Create application* on the applications page, enter the required and any optional details that you want to provide and confirm with *Create application* again. The *Create application* form also provides an option to customise your application manifest.
+Click *Create application* on the applications page, enter the required and any optional details that you want to provide and confirm with *Create application* again.
+
+An application can be created either in a VM or a container. Selecting one of these options will update the image field to only show images of the chosen type. The default image for the selected type is automatically populated in the image field.
+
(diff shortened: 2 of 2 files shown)


ajanon:
Is this for 1.25 too? If it is, please see my comment on your other PR: https://github.com/canonical/anbox-cloud-docs/pull/260#issuecomment-2582200754

anusha-c18:
This can be merged right away, it isn't targeted for 1.25 @ajanon

howto/application/delete-application.md
```diff
@@ -1,7 +1,7 @@
 (howto-delete-application)=
 # How to delete an application
 
-When an application is no longer needed, it can be fully removed from Anbox Cloud. Removing an application will cause all of its versions to be removed, including all of its currently active instances. Be extra careful as this might affect your users if any are still using instances of the application you want to remove.
+When an application is no longer needed, it can be fully removed from Anbox Cloud. Removing an application will cause all of its versions to be removed, including all of its currently active instances. However, the instance will first enter an error state before being cleaned up. Be extra careful as this might affect your users if any are still using instances of the application you want to remove.
```
ajanon, line 4: Is this supposed to happen?
@morphis @adglkh

adglkh: Just copy & paste I put on the mm change.
```
That's not the case so far, at least for the regular application.
E.g. if there are running instances, when their associated applications are  removed, they still remain active until people manually delete them.
I am under the impression the correct behavior that has been discussed about internally is that whenever an application gets deleted, its associated instance will be deleted as well regardless its base or regular.
However we're not there yet.
```
I will create a story tracking for this. 
The question is: do we want to mention this in our doc before the bug is addressed?

keirthana: In that case, I would leave out the sentence in question about the instance entering the error state since that is not expected behaviour and if users encounter it, they should file a bug with us. 

Let's stick to mentioning that when an application is deleted, its instances will also be deleted with the warning to be mindful of users still using the instances when an application is being deleted.

keirthana, line 4:
```suggestion
When an application is no longer needed, it can be fully removed from Anbox Cloud. However, be mindful of the consequences that come with removing an application such as:
- Removal of associated base instances. Note that the base instance could first enter an error state before being cleaned up.
- Removal of all the versions of the deleted application.
```

keirthana: @anusha-c18 A slight rewrite after the discussion in today's stand-up. Please see my suggestion below.

anusha-c18: Addressed the [changes](https://github.com/canonical/anbox-cloud-docs/compare/5c43e8e4920cdbce6f8782be278cbfd96d43ddf2..840dbbef714a527e69c0f003abc421a5cb90a7ce)

keirthana: Thanks Anusha, could you also rebase on `main` - I fixed some link check failures earlier today and I need the checks to pass to merge this change.

anusha-c18: Done :)

keirthana (approved):
LGTM

morphis (approved):
LGTM

PR #262, opened 2025-01-13
feat(instances): add docs for instance details page

anusha-c18:
Add information about the tabs in the instance details page.


keirthana (approved):
LGTM,Thanks!

PR #306, opened 2025-03-10
Fix infinite loop when running `make run`

keirthana:
# Documentation changes

When running `make run`, except for the very first time, an infinite loop runs for every doc change because of the auto-generated `ams-configuration.md` file. The script that generates this file writes the output irrespective of whether there are changes to that particular file or not. This file being generated irrespective of changes triggers sphinx autobuild which
builds again and again causing an infinite loop.

This commit fixes the infinite loop by writing to the `ams-configuration.md` file only if there are changes to it.

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

N/A

### DIFF
--- a/scripts/ams_configuration.py
+++ b/scripts/ams_configuration.py
@@ -2,6 +2,7 @@
 
 import argparse
 import json
+import os
 from typing import Dict
 
 import requests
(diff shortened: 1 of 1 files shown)


morphis (approved):
LGTM

PR #317, opened 2025-03-28
docs(charms): add note for bundle architectures

jat-canonical:
# Documentation changes

Add a note about the bundles being architecture dependent and the old overlay not working anymore.

# Review and preview

In your local repository,

## Reviewers

### DIFF
--- a/howto/install/deploy-juju.md
+++ b/howto/install/deploy-juju.md
@@ -271,16 +271,12 @@ Check the output of the `juju status` command to see whether you need to reboot:
 
 ```sh
 ...
-Unit       Workload  Agent  Machine  Public address  Ports  Message
-lxd/0*     active    idle   3        10.75.96.23            reboot required to activate new kernel
+Unit       Workload  Agent  Machine  Public address  Ports      Message
+lxd/0*     active    idle   3        10.75.96.23     8443/tcp   Actions: Reboot Required, Role: Database-leader
 ...
 ```
 To reboot the machine hosting LXD, run the following command:
 
     juju ssh lxd/0 -- sudo reboot
-
-When the machine is back running, you must manually clear the status of the LXD units:
-
-    juju run --wait=5m lxd/0 clear-notification
 
 Once done, the reboot operation is finished.
--- a/howto/install/deploy-juju.md
+++ b/howto/install/deploy-juju.md
@@ -246,9 +246,16 @@ machines:
     constraints: "instance-type=m6g.2xlarge root-disk=50G"
 ```
 
-To deploy, add `--overlay overlay.yaml` to your deploy command. For example:
+The default bundles for `anbox-cloud` and `anbox-cloud-core` available on Charmhub are suitable only for pure AMD64/X86 based deployments.
+Applying the overlay given above, to the bundle directly does not properly deploy the correct architecture for the charms.
+You need to download the bundle using,
 
-    juju deploy anbox-cloud --overlay ua.yaml --overlay overlay.yaml
+    juju download anbox-cloud --channel <bundle_channel>
+
+Unzip the bundle to extract the `bundle.yaml` file and remove the revisions for all Anbox Cloud charms from the bundle.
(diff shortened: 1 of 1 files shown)


howto/install/deploy-juju.md
```diff
@@ -250,6 +250,19 @@ To deploy, add `--overlay overlay.yaml` to your deploy command. For example:
 
     juju deploy anbox-cloud --overlay ua.yaml --overlay overlay.yaml
 
+```{important}
+The default bundles for `anbox-cloud` and `anbox-cloud-core` available on Charmhub are suitable only for pure AMD64/X86 based deployments.
+Applying the overlay given above, to the bundle directly does not properly deploy the correct architecture for the charms.
```
adglkh, line 255: Since you added the important notice under `# Use Arm instances`, and the bundle available on the store is only suitable for x86_64, I suggest dropping `{important}` notice, and moving the text added to the body in the context.
I feel much better in this way.

howto/install/deploy-juju.md
```diff
@@ -250,6 +250,19 @@ To deploy, add `--overlay overlay.yaml` to your deploy command. For example:
...
+You need to download the bundle using,
+
+    juju download anbox-cloud --channel <bundle_channel>
+
+Unzip the bundle to extract the `bundle.yaml` file and remove the all revisions from the bundle.
```
adglkh, line 260: Would the removal of revision extend to non-Anbox Cloud charms as well? E.g.
- haproxy `revision: 66` in the bundle while `revision: 84` in the latest/stable channel
- etcd `revison:774` in the bundle while `revision: 781` in the latest/stable channel
- easyrsa `revision:66` in bundle while `revision: 68` in the latest/stable channel

My concern here is that we don't test out those revisions in either integration tests or spread test yet.

howto/install/deploy-juju.md
jat-canonical: I was confused on the same. Let's see what @keirthana and @morphis think about it.

keirthana: I agree with Gary's suggestion.

keirthana (approved):
LGTM, if Gary's feedback is addressed.

howto/install/deploy-juju.md
jat-canonical: fixed this

howto/install/deploy-juju.md
```diff
@@ -250,6 +250,17 @@ To deploy, add `--overlay overlay.yaml` to your deploy command. For example:
 
     juju deploy anbox-cloud --overlay ua.yaml --overlay overlay.yaml
 
+The default bundles for `anbox-cloud` and `anbox-cloud-core` available on Charmhub are suitable only for pure AMD64/X86 based deployments.
+Applying the overlay given above, to the bundle directly does not properly deploy the correct architecture for the charms.
+You need to download the bundle using,
+
+    juju download anbox-cloud --channel <bundle_channel>
+
+Unzip the bundle to extract the `bundle.yaml` file and remove the revisions for all anbox cloud charms from the bundle.
```
keirthana, line 259:
```suggestion
Unzip the bundle to extract the `bundle.yaml` file and remove the revisions for all Anbox Cloud charms from the bundle.
```

keirthana: @jat-canonical I think with this small update, this is ready to go.

@adglkh Could you also have a look at this PR and leave your approval if you are good.

howto/install/deploy-juju.md
```diff
@@ -250,6 +250,17 @@ To deploy, add `--overlay overlay.yaml` to your deploy command. For example:
 
     juju deploy anbox-cloud --overlay ua.yaml --overlay overlay.yaml
```
adglkh, line 251: My suggestion is to remove the above command as it's mentioned in the following context since we know running the above command would fail under
```
Use Arm instances
```

jat-canonical: Makes sense, updated. Thanks.

ajanon (approved):
LGTM

adglkh (approved):
LGTM

PR #320, opened 2025-04-04
Rearrange security content

keirthana:
# Documentation changes

Based on the review by security team, this PR aims to rearrange the security content in such a way that component level security is highlighted in addition to the overall security posture of the product.

There is no change in technical aspects of what we are communicating as we have already reviewed those during our security documentation brainstorming sessions.

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

AC-3248


explanation/security/instance.md
```diff
@@ -0,0 +1,23 @@
+# Instances
+
+Anbox Cloud uses secure and isolated system instances supplied by [LXD](https://ubuntu.com/lxd). LXD provides a high degree of flexibility when setting up instances, allowing you to run in a fully secure or less secure way, depending on your requirements. See [Security](https://documentation.ubuntu.com/lxd/en/latest/security/) in the LXD documentation for more information about how a LXD setup can be secured.
+
+Using virtual machines to host Android containers provides better workload isolation. Thus, Anbox Cloud uses LXD in a way that enforces the highest security level.
```
ajanon, line 5: This is a bit unrelated to this PR as it is the same as before:
I find it a bit odd that the second sentence starts with `thus`. Most users don't run virtual machines, but system containers. While it's true that VMs provide better isolation, I don't really understand why this implies that 'Anbox Cloud uses LXD in a way that enforces the highest security level'. I don't have a better way to phrase this, however.

explanation/security/streaming-stack.md
```diff
@@ -0,0 +1,66 @@
...
+
+Anbox Streaming Gateway is using cryptographic technology for:
+
+* TLS transport encryption
+* Mutual TLS based authentication
+* Token based authentication
+
+### TLS transport encryption
+
+All network endpoints exposed by the Anbox Stream Gateway are secured with TLS using an 4096 bit RSA key. The Anbox Stream Gateway strictly enforces TLS 1.3 or later and does not provide backward compatibility with older TLS versions.
```
ajanon, line 23: I think this should be _a_ 4096 as 4096 starts with a consonant. Wdyt?
```suggestion
All network endpoints exposed by the Anbox Stream Gateway are secured with TLS using a 4096 bit RSA key. The Anbox Stream Gateway strictly enforces TLS 1.3 or later and does not provide backward compatibility with older TLS versions.
```

ajanon (changes requested):
A few nitpicks, but looks good to me overall

explanation/security/instance.md
keirthana: Yes, I didn't really focus on correcting the legacy content but nice catch.

I have tried to make the two paragraphs read logically better. How does it sound to you now?

ajanon (approved):
LGTM, thanks!

morphis (approved):
LGTM

Issue #214, opened 2024-11-12
Consider splitting upgrade guides by version


lathiat:
Currently, the Anbox docs only seem to render a "latest" version. As such, the upgrade guide shows you how to upgrade the previous version to the current release, and the commands currently have the 1.23 channel specified. This makes it harder to know how to upgrade 1.21 to 1.22 if you haven't yet upgraded.

Frequently customers will end up behind versions, and may need to refer to an upgrade guide for a previous versions. I am not sure how much we expect customers to end up behind on upgrades with Anbox, but with other products this is very common, but I imagine at least it may be more common for people to end up 2 versions behind if they didn't quite finish upgrades before the next release came out.

Some approaches that may work:
- Some products let you view the older releases documentation in full, 
- Some products have a separate upgrade page for each version, e.g. 1.20 to 1.21, 1.21 to 1.22, 1.22 to 1.23, etc
- Keeping a single latest "upgrade" page -might- make sense but we need to make sure that the upgrade page makes very clear how to check the current version, if you can skip a version or not, and maintains any release-specific exceptional upgrade notes in a clearly marked section (we have a couple of those in the anbox upgrade docs at the moment)

Our Kubernetes docs used to have a simple list in the documentation tree for each version as noted above, however that seems to have been lost at some point and now instead links to the "per version" upgrade page, although our docs have no other way to switch between versions.. normally if you do this there should be a clear selector on every docs page to jump between versions of the page, and we seem to have lost that. In any case, it is a bit of an example: https://ubuntu.com/kubernetes/docs/1.23/upgrading

lathiat:
MySQL has a great example of linking between versions, with a version selector on all pages in the top right:
https://dev.mysql.com/doc/refman/9.1/en/upgrading.html

keirthana:
We understand the issue but the challenge lies in maintaining documentation only for the supported versions. So out of the options you have listed the last one about keeping a single page but having some more information that will help the upgrade process for older versions is the one that makes most sense. 

I will discuss this with the team and come up with a Jira ticket to get this fixed.

keirthana:
https://warthogs.atlassian.net/browse/AC-3141 is created to fix this. We won't be versioning documentation yet as officially we support only the latest version. However, the documentation can be fixed to include guidelines on how to check for the latest version that a particular user is on and then advise on how to get to a later version from there.

keirthana:
[...]

PR #327, opened 2025-04-22
Clarify upgrade documentation

keirthana:
# Documentation changes

This PR clarifies the upgrade documentation to state that we don't officially support any other upgrade than from the n-1 to n upgrade path. This should fix issue #214. 

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

AC-3141

### DIFF
--- a/howto/upgrade/upgrade-anbox.md
+++ b/howto/upgrade/upgrade-anbox.md
@@ -5,9 +5,9 @@
 If you're interested in getting notified for the latest Anbox Cloud releases, make sure you subscribe to the [Anbox Cloud category](https://discourse.ubuntu.com/c/anbox-cloud/49) on the Ubuntu discourse.
 ```
 
-Anbox Cloud officially supports only the latest release. This means that we support upgrades from n-1 to nth minor version, where n is the most recent minor version released.
+Anbox Cloud supports only the latest release. This means that we support upgrades from n-1 to nth minor version, where n is the most recent minor version released.
 
-If you are on an earlier version of Anbox Cloud than the most recent one, we recommend upgrading to the latest version as soon as possible. If you are multiple versions behind the supported version, there could be potential disruptions if you upgrade directly to the current version. You should keep this in consideration while planning your upgrade.
+If you are on an earlier version of Anbox Cloud than the most recent one, we recommend upgrading to the latest version as soon as possible. If you are multiple versions behind the supported version, there could be potential disruptions if you upgrade directly to the current version. The best practice in such a scenario is to upgrade incrementally, version by version.
 
 This guide contains upgrade instructions for the charmed Anbox Cloud deployment. In addition to the upgrade of the charms, any used images or addons need to be updated as well.
 
@@ -157,7 +157,7 @@ If you are running LXD clusters with the LXD snap from a channel other than 5.21
 
     juju config lxd channel=5.0/stable
 
-where `5.0/stable` is the current LXD channel. Not doing this might lead to a broken LXD cluster.
+where `5.0/stable` is the currently installed LXD snap channel. Not doing this might lead to a broken LXD cluster.
 ```
 
 To start, upgrade the AMS LXD charm to the latest revision using:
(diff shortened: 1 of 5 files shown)


howto/upgrade/upgrade-anbox.md
```diff
@@ -5,23 +5,28 @@
 If you're interested in getting notified for the latest Anbox Cloud releases, make sure you subscribe to the [Anbox Cloud category](https://discourse.ubuntu.com/c/anbox-cloud/49) on the Ubuntu discourse.
 ```
 
-Anbox Cloud allows upgrades from older versions to newer version. This describes the steps necessary to perform the upgrade.
+Anbox Cloud officially supports only the latest release. This means that we support upgrades from n-1 to nth minor version, where n is the most recent minor version released.
```
ajanon, line 8: Should we give an example to clarify? e.g.:
```suggestion
Anbox Cloud officially supports only the latest release. This means that we support upgrades from n-1 to nth minor version, where n is the most recent minor version released. For example, we only support upgrades from Anbox Cloud 1.24 to Anbox Cloud 1.25.
```

I'm not sure if it's necessary. Without the example, I feel like there should be one, but adding the example feels a bit silly given that it's very simple.
Wdyt?

howto/upgrade/upgrade-anbox.md
```diff
@@ -5,23 +5,28 @@
...
 
-The upgrade instructions detail the revisions each charm needs to be upgraded to, to bring it to the latest version. Next to the upgrade of the charms any used images or addons need to be updated as well.
+If you are on an earlier version of Anbox Cloud than the most recent one, we recommend upgrading to the latest version as soon as possible. If you are multiple versions behind the supported version, there could be potential disruptions if you upgrade directly to the current version. You should keep this in consideration while planning your upgrade.
```
adglkh, line 10: Maybe we could also mention that `if an existing deployment is several versions behind the most recent one, the best practice is to upgrade incrementally, version by version.`, what do you think?

howto/upgrade/upgrade-anbox.md
keirthana: I don't want to give the example because it can get outdated as we move ahead. Agreed, it is only an example but when we say n is the latest version and when the next statement gives an example of older versions, it doesn't process well in perception.

howto/upgrade/upgrade-anbox.md
```diff
@@ -118,21 +112,21 @@ Since the NATS charm has been overhauled to use the modern charm framework (Ops
...
+As the last step, upgrade the LXD cluster. Upgrading LXD will not restart running instances but it's recommended to take a backup before continuing.
 
 ```{note}
 In Anbox Cloud's 1.25.0 release, the LXD charm has been reworked and hence there are changes in the upgrade instructions for the charm.
 ```
 
 <details>
 <summary>Click here for upgrade instructions for Anbox Cloud versions < 1.25.0 </summary>
-In some cases, specifically when you maintain bigger LXD clusters or want to keep a specific set of LXD nodes active until users have dropped, it makes sense to run the upgrade process manually on a per node basis. To enable this, you can set the following configuration option for the LXD charm before running the refresh command above:
+In some cases, specifically when you maintain bigger LXD clusters or want to keep a specific set of LXD nodes active until users have dropped, it is pragmatic to run the upgrade process manually on a per node basis. To manually upgrade individual nodes, run the following command before running the `juju refresh` command:
```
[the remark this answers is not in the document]
keirthana: Agreed that this is outdated. But this is inside a collapsible drop down for upgrades from versions earlier than 1.25.0. Since 1.25.0 is the current release and there is a possibility that we have customers on unsupported versions, I'd like to keep this for a little while more, at least until 1.27.0 and then remove it.

howto/upgrade/upgrade-anbox.md
```diff
@@ -146,51 +140,49 @@ Once the unnecessary nodes are dropped, the upgrade for a single LXD deployment
 
 Once the upgrade has completed, the unit will be marked as active.
 
-For major and minor version upgrades, an update of the LXD charm may upgrade kernel modules or GPU drivers. This requires stopping any running instances before applying the upgrade and performing a reboot of the machine once the upgrade completed.
+Major and minor version upgrades may include upgrades to kernel modules or GPU drivers. This requires stopping any running instances before applying the upgrade and performing a reboot of the machine once the upgrade completed.
 
-In case a reboot of the machine is required, a status message will be shown. When the machine has been rebooted, the status message can be cleared by running:
+In case a reboot is required, a notification is displayed. When the machine is rebooted, you can clear the notification by running:
```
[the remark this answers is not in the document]
keirthana: see above

jat-canonical (approved):
LGTM

howto/upgrade/upgrade-anbox.md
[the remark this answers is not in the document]
adglkh: That makes sense to me.
Thanks

reference/release-notes/release-notes.md
```diff
@@ -35,7 +35,7 @@ Minor releases
 Patch releases
 : A patch release for Anbox Cloud is released every month and includes Android and Chrome security updates alongside Anbox Cloud specific bug fixes.
 
-Anbox Cloud currently officially supports only the most recent release. Older releases are only supported for a short time after a new minor release was published.
+Anbox Cloud officially supports only the most recent release. Older releases are only supported for a short time after a new minor release was published.
```
adglkh, line 38: The term "officially" is also mentioned here.

adglkh (approved):
LGTM. One nitpick though

howto/upgrade/upgrade-anbox.md
ajanon: Makes sense.

PR #332, opened 2025-04-29
Add constraint for scaling down cluster

### LINKED ISSUE #331: Missing information about scaling down
When scaling down the LXD cluster, a DB leader node should not be removed.

Relevant information from [MM conversation](https://chat.canonical.com/canonical/pl/r7w7xbcwx38xmm5k5oe31r93se):

Removing the db-leader node specifically is not recommended but removing the db nodes should be fine since lxd would choose other nodes as db nodes in that case afair. Generally you should avoid touching any of the DB nodes unless you remove everything as removing them includes a risk of things to break.

keirthana:
# Documentation changes

Add constraint for scaling down a LXD cluster about how database leader node should not be removed.

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

Fixes #331

### DIFF
--- a/howto/cluster/scale-down.md
+++ b/howto/cluster/scale-down.md
@@ -3,18 +3,18 @@
 
 Scaling down a LXD cluster involves more checks than scaling up.
 
-There are two important requirements when scaling down:
+### Prerequisites
+
+The following are important requirements when scaling down:
  - The node you remove must not have any instances left.
  - You must wait for a node to be fully removed before you can start removing another one.
-
-```{important}
-Since Anbox Cloud 1.25.0, the default channel for the LXD charm has changed to 5.21/stable.
+ - Unless you are removing the entire cluster or are prepared for breaking changes, you should not remove database nodes. Especially, never remove a database leader node unless you are planning to remove the entire cluster.
+- Since Anbox Cloud 1.25.0, the default channel for the LXD charm has changed to 5.21/stable.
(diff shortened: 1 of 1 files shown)


howto/cluster/scale-down.md
```diff
@@ -3,18 +3,18 @@
...
-There are two important requirements when scaling down:
+### Prerequisites
+
+The following are important requirements when scaling down:
  - The node you remove must not have any instances left.
  - You must wait for a node to be fully removed before you can start removing another one.
-
-```{important}
-Since Anbox Cloud 1.25.0, the default channel for the LXD charm has changed to 5.21/stable.
+ - Unless you are removing the entire cluster or are prepared for breaking changes, you should not remove database nodes. Especially, never remove a database leader node unless you are planning to remove the entire cluster.
```
adglkh, line 11: For average people, how do they know which lxd nodes are database nodes or leader node?

keirthana: Is there a command to identify or should the users remember which node they set as leader @jat-canonical ?

keirthana: Will `lxc cluster list` show?

adglkh: In the context of removing an LXD unit via juju comment, we need to clarify how to determine whether the underlying lxd node is a database node or the cluster leader. This is important because the role shown from juju perspective does not necessarily align with the actual role of the node within the LXD cluster.

For example, a unit marked as the leader in the output of `juju status might` be running on:
- leader
- database node
- on-database node

All of these are possible.

keirthana: I added how to identify from the LXD perspective because in case the mismatch happens between Juju and LXD, the situation is already out of control. Here, we are providing a prerequisite for scaling down, so it is enough to provide how to identify from the LXD perspective and not remove those nodes.

PR #373, opened 2025-06-12
Clarify GPU slots usage

keirthana:
# Documentation changes

This PR clarifies documentation around our GPU slots and GPU encoder slots assignment. The default values documented per GPU vendor will be corrected via https://github.com/canonical/ams/pull/1227

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

AC-3466

### DIFF
--- a/explanation/capacity-planning.md
+++ b/explanation/capacity-planning.md
@@ -56,7 +56,7 @@ When you launch an instance for an application, AMS reserves the number of GPU s
 ```{important}
 GPU slots are used to share GPUs among instances, but they do not impose limits on GPU usage. Therefore, increasing the number of required GPU slots for an application does not guarantee that more GPU resources are allocated to the corresponding application instances. For example, an intensive game that is configured to use one GPU slot might consume more GPU resources than a simple photo gallery app that is configured to use five GPU slots.
 
-The main purpose of GPU slots is to control the number of instances that are launched on a node that has a GPU installed, which reduces contention for GPU resources.
+There is a default number of GPU slots and GPU encoder slots configured depending on the GPU used but this can be changed. The main purpose of GPU slots is to control the number of instances that are launched on a node that has a GPU installed, which reduces contention for GPU resources.
 ```
 (sec-over-committing)=
 ## Over-committing resources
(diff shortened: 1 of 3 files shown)


ajanon (approved):
LGTM, thanks!

explanation/capacity-planning.md
```diff
@@ -56,7 +56,7 @@ When you launch an instance for an application, AMS reserves the number of GPU s
 ```{important}
 GPU slots are used to share GPUs among instances, but they do not impose limits on GPU usage. Therefore, increasing the number of required GPU slots for an application does not guarantee that more GPU resources are allocated to the corresponding application instances. For example, an intensive game that is configured to use one GPU slot might consume more GPU resources than a simple photo gallery app that is configured to use five GPU slots.
 
-The main purpose of GPU slots is to control the number of instances that are launched on a node that has a GPU installed, which reduces contention for GPU resources.
+There is a default number of GPU slots and GPU encoder slots configured depending on the GPU used but this can be changed. The main purpose of GPU slots is to control the number of instances that are launched on a node that has a GPU installed, which reduces contention for GPU resources.
```
adglkh, line 59: If I were a user, I might wonder what the default number of GPU slots and GPU encoder slots is for a given GPU.
Maybe we could suggest using `amc node show <node>` to query this information.

howto/cluster/configure-nodes.md
```diff
@@ -61,6 +61,8 @@ Use the following command to prevent the node from accepting new instances:
 
 GPU slots are used to share GPUs among instances. Each GPU-equipped cluster node is configured with a number of GPU slots and a number of GPU encoder slots. See {ref}`sec-node-configuration` for the default values that are used. Nodes without GPU are configured with 0 GPU slots and 0 GPU encoder slots.
 
+The number of GPU slots assigned per node is arbitrary and can be changed when creating an application or an instance. It is used only to manage and assign hardware resources. When using GPU encoding, the number of GPU slots per node is limited by the number of GPU encoder slots, i.e., the number of GPU slots cannot exceed the number of GPU encoder slots. When using software encoding, GPU slots can be more than the GPU encoder slots.
```
adglkh, line 64: The first sentence confused me a bit, especially for the first part. What you want to explain to me is more like `The number of available GPU slots per node varies depending on the number of running GPU-based anbox instances and can be changed when .......` .

Meanwhile, I don't fully the sentence `When using GPU encoding, the number ... " could you explain a bit more to me?

explanation/capacity-planning.md
keirthana: It is fair to expect to know what is the default number but this is an explanation where users are trying to understand the meaning behind these configuration.

So I have linked the configuration topic that will list the defaults.

howto/cluster/configure-nodes.md
adglkh: > We want to convey that during GPU encoding, the GPU slots value cannot exceed the GPU encoder slots value but while software encoding, it can.


A: Okay, is this per node, per application, or per instance?

explanation/capacity-planning.md
adglkh: Okay, that makes sense to me.

adglkh (approved):
LGTM

PR #381, opened 2025-06-25
Rewrite access Android instance topic

keirthana:
# Documentation changes

- Rewrite secure connection via Anbox Connect
- Remove insecure way of exposing ADB service
- Move share session and update share instructions to a separate guide
- Add necessary redirects

Context: The motivation for this PR came from the exercise of preparing the user discussion guide for the Android track. When doing that, I realized due to the timeline of when we implemented Anbox Connect and what we offered before, the current how-to guide is very confusing for users, mixing multiple concepts. It also offered less than ideal ways (insecure connection using ADB) when we offer a clearly better and secure alternative way.

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

AC-3513


howto/android/access-android-instance.md
```diff
@@ -0,0 +1,65 @@
...
+
+After verifying that the instance is running and has a session ID, share the session:
+
+    sudo anbox-cloud-appliance.gateway session share <session_id> --description="remote access to demo instance"
+
+```{tip}
+Providing a description helps you identify a shared session when you are sharing a session multiple times with different people. 
+```
+
+The output returns the command you need to run next from your host machine: `anbox-connect <connection_url>`
```
adglkh, line 48: returns the command
->
returns a URL

howto/instance/share-session.md
```diff
@@ -0,0 +1,38 @@
+(howto-share-session)=
+# Share a session
+
+A session in Anbox Cloud denotes the streaming session of an Anbox instance. Each session of an instance can be shared up to a maximum of 5 times. By default, each share will expire after 12 hours.
```
adglkh, line 4: When a presigned URL is generated from the dashboard, it’s only valid for 6 hours
image

adglkh (approved):
Two minor ones, otherwise LGTM

howto/android/access-android-instance.md
keirthana: This actually returns anbox-connect <URL> which we can copy as is and run. That's why I wrote command.

howto/instance/share-session.md
keirthana: Good catch, why is there a difference between the CLI and the dashboard? If there should be feature parity, I can file a bug.

howto/android/access-android-instance.md
adglkh: Ah, okay. I missed the context. 
Thanks for the correction.

howto/instance/share-session.md
adglkh: Please do.
Looks like we forgot to update the default expiration time for the session sharing URL in the web UI after we had agreed to increase it to 12 hours during one of our weekly sync meetings.
See [Default expiration time of a share](https://docs.google.com/document/d/1aLOMrzKvBjpqg6arvrW1lJ0_Fh1T5JaP7F1jKTkcAnI/edit?tab=t.0)

keirthana: I can see that this is already fixed in 1.27
image

lorumic: > When a presigned URL is generated from the dashboard, it’s only valid for 6 hours

Maybe you opened the modal 6 hours after generating it? It has never been set to a default 6 hours validity in the dashboard.

keirthana: Nice! I am impressed by the amount of thought that has gone into when to enable the validity button and showing how much time is left. Thank you Michele! 

@adglkh If you are good with this, I can merge the PR as is.

adglkh: Okay, thank you both for the verification.

PR #389, opened 2025-07-08
feat: add docs for audience url while using auth0

anusha-c18:
# Documentation changes

- add audience url for when the access token will be used 

# Review and preview

In your local repository,

## Reviewers

### DIFF
--- a/howto/install-appliance/setup-oidc/auth0.md
+++ b/howto/install-appliance/setup-oidc/auth0.md
@@ -18,8 +18,9 @@ Auth0 is a flexible, drop-in solution to add authentication services to your app
     - Enable *Allow Refresh Token Rotation*.
     - Select *Save Changes*.
 
-1. Near the top of the *Settings* tab, locate the *Domain* field. Copy the value and prefix it with `https://`. This is your issuer URL.
+1. Copy the following values for the preseed YAML configuration:
+    - Near the top of the *Settings* tab, locate the *Domain* field, and prefix it with `https://`. This is your issuer URL.
+    - From the *Settings* tab, copy the *Client ID*.
+    - Select *Applications > APIs* and copy the *API Audience* value.
 
-1. From the *Settings* tab and copy the *Client ID*.
-
-1. {ref}`Configure the preseed YAML <howto-configure-oidc>` with the client ID and issuer URL values obtained in the previous steps.
+1. {ref}`Configure the preseed YAML <howto-configure-oidc>` with the issuer URL, client ID, and API audience URL values obtained in the previous steps.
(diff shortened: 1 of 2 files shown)


jat-canonical (approved):
LGTM. Thanks.

howto/install-appliance/setup-oidc/auth0.md
```diff
@@ -22,4 +22,6 @@ Auth0 is a flexible, drop-in solution to add authentication services to your app
 
 1. From the *Settings* tab and copy the *Client ID*.
 
-1. {ref}`Configure the preseed YAML <howto-configure-oidc>` with the client ID and issuer URL values obtained in the previous steps.
+1. In the main navigation, select *Applications > APIs*. Copy the *API Audience* value.
```
keirthana, line 25:
```suggestion
1. Select *Applications > APIs* and copy the *API Audience* value.
```
It would also be good to add a line about why we are copying this value. Something like - "This value will be used when you do xxx"

keirthana (changes requested):
One minor suggestion

howto/install-appliance/setup-oidc/auth0.md
anusha-c18: > Let's try to avoid documenting every UI element, the idea is to document the goals of the user and not to document the UI as they see it. Most user interfaces are intuitive and self-explanatory.

The applications page form has several tabs and accordions. All the steps so far have been within the same page. However, this step requires the user to go to a new page. I think it’s worth keeping it this way so that users don’t look for a tab/accordion within the applications page. 

> It would also be good to add a line about why we are copying this value. Something like - "This value will be used when you configure the preseed YAML"

This is added in the next point. Should we add it to this point as well?

keirthana: > The applications page form has several tabs and accordions. All the steps so far have been within the same page. However, this step requires the user to go to a new page. I think it’s worth keeping it this way so that users don’t look for a tab/accordion within the applications page.

I think it is counter intuitive for someone to search for Applications as an accordion when they are within the applications page. (Since we are still saying Applications > APIs)


> This is added in the next point. Should we add it to this point as well?

I would recommend providing the context here than in the next point. Users prefer to have the context when they are doing some action. So instead of asking them to do what we say mechanically, something like this may work better (this approach may be required to be applied globally):

Select *Applications > APIs*. Copy the *API Audience* value, you will need this when you configure the preseed YAML.

Now that you have the client ID, API audience URL and issuer URL, configure the preseed YAML.

anusha-c18: Makes sense, addressed the [changes](https://github.com/canonical/anbox-cloud-docs/compare/68532054fe8702fc6e39f669cb8a8850b34d6264..4a733bb31303e1382c5dd1c6e1121023690a01b3)

keirthana (approved):

PR #419, opened 2025-08-06
Recommend Juju 3.6 and list Juju 3.0 as minimum version

keirthana:
# Documentation changes

This PR drops Juju 2.9 support and adds Juju 3.0 as the minimum required version and Juju 3.6 as the recommended version.

# Review and preview

Yes

## Reviewers

# JIRA / Launchpad bug

AC-3552

### DIFF
--- a/reference/requirements.md
+++ b/reference/requirements.md
@@ -79,15 +79,17 @@ Charmed Anbox Cloud requires LXD version >= 5.0.
 (sec-juju-version-requirements)=
 ### Juju
 
-The charmed Anbox Cloud requires [Juju 2.9 or later](https://juju.is/) to manage the different components and their dependencies. You can install Juju with the following command:
+The charmed Anbox Cloud requires a minimum of [Juju 3.0 or later](https://juju.is/) to manage the different components and their dependencies.
 
-    snap install --classic --channel=3.1/stable juju
+If you want to manage user secrets, use Juju 3.3 or later.
 
-If you wish to install a different version, replace `3.1` in the command with the desired version.
+We recommend using Juju 3.6, so that you will be using the latest version of Juju and can manage user secrets. See [Juju documentation](https://documentation.ubuntu.com/juju/latest/howto/manage-secrets/) for more information about managing secrets.
 
-To switch to the 2.9 series, use the following command:
+You can install Juju with the following command:
 
-    snap refresh --channel=2.9/stable juju
+    snap install --classic --channel=3.6/stable juju
+
+If you wish to install a different version, replace `3.6` in the command with the desired version.
 
 See the [Juju documentation](https://canonical-juju.readthedocs-hosted.com/en/latest/user/howto/manage-juju/#install-juju) for more information.
 
--- a/explanation/security/charms.md
+++ b/explanation/security/charms.md
@@ -37,9 +37,6 @@ juju list-secrets
 ```
 
 Currently, all Juju secrets are managed by the charms themselves, not by the model. This means you can view them, but you cannot rotate or remove them.
-```{note}
-Juju secrets are disabled in Anbox Cloud deployments based on Juju 2.9, as the secret feature is only available starting from Juju 3.0.
-```
 
 ### Use Juju secrets across model
 
--- a/howto/upgrade/upgrade-anbox.md
+++ b/howto/upgrade/upgrade-anbox.md
@@ -24,17 +24,6 @@ Read the release notes for your target version to learn about important changes 
 Check the [required Juju version](https://documentation.ubuntu.com/anbox-cloud/reference/requirements/#juju).
 If your deployment uses an earlier Juju version, you must upgrade your controller and all models first. See the [Juju documentation](https://canonical-juju.readthedocs-hosted.com/en/latest/user/howto/manage-models/#upgrade-a-model) for instructions on how to upgrade the Juju controller and all models to a newer Juju version.
 
-```{note}
-
-The following instructions assume that you're using Juju >= 3.1. If you're using Juju 2.9, you have to map the following commands:
-
-| Juju 3.x | Juju 2.9 |
-|----------|----------|
-| `juju refresh` | `juju upgrade-charm` |
-| `juju exec` | `juju run` |
-
-```
-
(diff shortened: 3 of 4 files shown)


reference/requirements.md
```diff
@@ -79,15 +79,17 @@ Charmed Anbox Cloud requires LXD version >= 5.0.
...
 ### Juju
 
-The charmed Anbox Cloud requires [Juju 2.9 or later](https://juju.is/) to manage the different components and their dependencies. You can install Juju with the following command:
+The charmed Anbox Cloud requires a minimum of [Juju 3.0 or later](https://juju.is/) to manage the different components and their dependencies.
 
-    snap install --classic --channel=3.1/stable juju
+If you want to manage user secrets, use Juju 3.3 or later.
 
-If you wish to install a different version, replace `3.1` in the command with the desired version.
+We recommend using Juju 3.6, so that you will be using the latest version of Juju and can manage user secrets. See [Juju documentation](https://documentation.ubuntu.com/juju/latest/howto/manage-secrets/) for more information about managing secrets.
```
jat-canonical, line 86: Should we add that it is the LTS release and has the full feature set that anbox cloud charms support.

reference/requirements.md
```diff
@@ -79,15 +79,17 @@ Charmed Anbox Cloud requires LXD version >= 5.0.
 (sec-juju-version-requirements)=
 ### Juju
 
-The charmed Anbox Cloud requires [Juju 2.9 or later](https://juju.is/) to manage the different components and their dependencies. You can install Juju with the following command:
+The charmed Anbox Cloud requires a minimum of [Juju 3.0 or later](https://juju.is/) to manage the different components and their dependencies.
 
-    snap install --classic --channel=3.1/stable juju
+If you want to manage user secrets, use Juju 3.3 or later.
```
jat-canonical, line 84: Should we rather mention that anbox cloud charms support user secrets which are only available in 3.3 or later ?

jat-canonical (commented):
two nits. Otherwise LGTM.

reference/requirements.md
```diff
@@ -79,15 +79,17 @@ Charmed Anbox Cloud requires LXD version >= 5.0.
 (sec-juju-version-requirements)=
 ### Juju
 
-The charmed Anbox Cloud requires [Juju 2.9 or later](https://juju.is/) to manage the different components and their dependencies. You can install Juju with the following command:
+The charmed Anbox Cloud requires a minimum of [Juju 3.0 or later](https://juju.is/) to manage the different components and their dependencies.
```
adglkh, line 82: This should be Juju 3.3 as the minimal requirement version for anbox cloud charm 1.28 to function

reference/requirements.md
```diff
@@ -79,15 +79,17 @@ Charmed Anbox Cloud requires LXD version >= 5.0.
...
+If you want to manage user secrets, use Juju 3.3 or later.
 
-If you wish to install a different version, replace `3.1` in the command with the desired version.
+We recommend using Juju 3.6, so that you will be using the latest version of Juju and can manage user secrets. See [Juju documentation](https://documentation.ubuntu.com/juju/latest/howto/manage-secrets/) for more information about managing secrets.
 
-To switch to the 2.9 series, use the following command:
+You can install Juju with the following command:
 
-    snap refresh --channel=2.9/stable juju
+    snap install --classic --channel=3.6/stable juju
```
adglkh, line 90: We should not add `--classic` to install juju from 3.6/ stable channel as it's now strit confinement

adglkh (changes requested):

reference/requirements.md
adglkh: Also, we're using channel `3/stable` in all other places. So for consistency, we may want to use 3/stable channel as well

reference/requirements.md
keirthana: Technically, someone can use Juju 3.0 and not use the "user secrets" feature right? A minimum requirement is something users need for Anbox Cloud to function and be fully supported by us. We agreed in a discussion during standup that 3.0 should be minimum version and for user secrets to be used, people need 3.3 (but we recommend 3.6 since that is the latest LTS)

Do you mean to say that 1.27 deployments will be rendered non functional if users have Juju less than 3.3?

adglkh: > Technically, someone can use Juju 3.0 and not use the "user secrets" feature right? A minimum requirement is something users need for Anbox Cloud to function and be fully supported by us. We agreed in a discussion during standup that 3.0 should be minimum version and for user secrets to be used, people need 3.3 (but we recommend 3.6 since that is the latest LTS)

A: yes, you're right

> Do you mean to say that 1.27 deployments will be rendered non functional if users have Juju less than 3.3?

A: No, using Juju versions lower than 3.3 is fine for 1.27 as we haven't dropped the compatibility layer in our charm code.

I mixed up this page with 
https://documentation.ubuntu.com/anbox-cloud/reference/deprecation-notices/
Sorry, 

Given that this PR targets the next-release branch for the upcoming 1.27.0 release, should we update the deprecation-notices page to indicate that Juju 3.3 will be the minimum required version for Anbox Cloud 1.28.0?

jat-canonical (approved):
LGTM

reference/requirements.md
```diff
@@ -79,15 +79,17 @@ Charmed Anbox Cloud requires LXD version >= 5.0.
...
+Anbox Cloud charms support user secrets and if you would like to use that feature, use Juju 3.3 or later.
 
-If you wish to install a different version, replace `3.1` in the command with the desired version.
+We recommend using Juju 3.6, so that you will be using the latest LTS version of Juju and can manage user secrets. Juju 3.6 has the full range of features that Anbox Cloud charms support. See [Juju documentation](https://documentation.ubuntu.com/juju/latest/howto/manage-secrets/) for more information about managing secrets.
 
-To switch to the 2.9 series, use the following command:
+You can install Juju with the following command:
 
-    snap refresh --channel=2.9/stable juju
+    snap install --channel=3.0/stable juju
```
adglkh, line 90: Basically, there is no 3.0/stable channel for juju
Let's keep things consistent and use 3/stable everywhere.

PR #471, opened 2025-09-23
Command reference and API specs workflow rework

ajanon:
# Documentation changes

Small improvements to the command reference and API specs workflow.

* Small refactors to make both workflows consistent.
* The update command reference workflow uses latest/stable instead of what's in .base_version
* The PR and commit messages for the command reference update workflow now show the appliance version
* Branches for PRs now contain the version of the appliance used
* Both workflows run weekly on Thursdays and Fridays, only after the tenth of each month. As we usually release on Wednesday in the middle of the month, this should avoid superfluous runs, but should get us runs when we need them. We can still trigger these manually as necessary.

# Review and preview

N/A

## Reviewers

N/A

# JIRA / Launchpad bug

N/A


adglkh (approved):
LGTM

PR #612, opened 2026-04-17
Created 1.30.0 release notes. Also added metadata to all existing pages.

bboisseau:
# Documentation changes

[Summary of documentation updates]

# Review and preview

In your local repository,

## Reviewers

# JIRA / Launchpad bug

### DIFF
--- a/scripts/add_descriptions.py
+++ b/scripts/add_descriptions.py
@@ -1,0 +1,481 @@
+#!/usr/bin/env python3
+"""
+Add HTML meta descriptions to Anbox Cloud documentation source files.
+
+Each page gets a 120-160 character plain-text summary injected into its
+MyST YAML frontmatter under the ``html_meta.description`` key, which
+Sphinx renders as ``<meta name="description" content="...">``.
+
+Usage
+-----
+    # List discovered docs files (bash one-liner equivalent)
+    python3 scripts/add_descriptions.py --list
+
+    # Dry-run: print what would change without writing files
+    python3 scripts/add_descriptions.py --dry-run
+
+    # Write descriptions to files that do not yet have one
+    python3 scripts/add_descriptions.py --write
+
+    # Overwrite existing descriptions too
+    python3 scripts/add_descriptions.py --write --overwrite
+
+    # Check that every file has a valid description (use in CI)
+    python3 scripts/add_descriptions.py --check
+
+    # Process a single file
+    python3 scripts/add_descriptions.py --write --file explanation/aaos.md
+
+Makefile shortcut: ``make descriptions`` / ``make check-descriptions``
+"""
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from pathlib import Path
+
+try:
+    import yaml
+    HAS_YAML = True
+except ImportError:
+    HAS_YAML = False
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+REPO_ROOT = Path(__file__).parent.parent
+
+# Directories (relative to REPO_ROOT) excluded from processing.
+# - reference/cmd-ref  → auto-generated; would be clobbered on next run
+# - reference/release-notes → factual records, not conceptual content pages
+EXCLUDED_PATHS: list[str] = [
+    ".sphinx",
+    "_build",
+    "_static",
+    "_templates",
+    "scripts",
+    "reuse",
+    ".github",
+    "images",
+    ".pytest_cache",
+    # Auto-generated from tool source code; descriptions would be overwritten.
+    "reference/cmd-ref",
+    # Template file only — not a published page.
+    "reference/release-notes/release-notes-template.md",
+]
+
+# Target character range for descriptions.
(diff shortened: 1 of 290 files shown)


ajanon:
What is this about, actually? Why do we need metadata to all pages?

bboisseau:
@ajanon this is requested for SEO, it will also allow us to improve our documentation in the long run.

scripts/add_descriptions.py
```diff
@@ -0,0 +1,481 @@
...
+            cut = raw.rfind(" ", 0, DESC_MAX)
+        desc = raw[:cut if cut > 0 else DESC_MAX]
+
+    desc = desc.rstrip(".,;:")
+    if desc and desc[-1] not in ".!?":
+        desc += "."
+
+    # Accept descriptions that are shorter than DESC_MIN when content is sparse
+    # (they are still useful even if below the target length).
+    return desc
```
ajanon, line 254: This is extremely complicated. Why not use a dedicated markdown parser? There are many of these in Python.

As-is, this is super hard to understand.

ajanon (commented):
One first comment. Also, I'm entirely unconvinced by the approach here, tbh.

I think this should be manual work, because as-is, this is not giving great results.

Just looking at a few of them:
* explanation/security/ams.md
> AMS. Anbox Management Service (AMS) is using cryptographic technology for: All network endpoints exposed by the AMS server are secured with TLS using an 4096.
* explanation/security/anbox-runtime.md
> Anbox runtime. The Anbox runtime (see ) is using cryptographic technology for: All HTTP communication Anbox performs over IP connections to external services.
* explanation/web-dashboard.md
> Anbox Cloud dashboard. The Anbox Cloud dashboard offers a web interface to create, manage and stream applications. The dashboard is especially helpful for
* howto/troubleshoot/troubleshoot-streaming-issues.md
> Troubleshoot streaming issues. Debugging issues that occur when streaming your application can be tricky. The following instructions give you some pointers on


Granted, I'm not entirely sure how SEO works, but having incomplete sentences does not look great, IMO.

We can always polish these manually after the fact, but I wonder what's the benefit of using an automated script if we have to manually fix it afterwards?

bboisseau:
> One first comment. Also, I'm entirely unconvinced by the approach here, tbh.
> 
> I think this should be manual work, because as-is, this is not giving great results.
> 
> Granted, I'm not entirely sure how SEO works, but having incomplete sentences does not look great, IMO.

Well spotted, I can fix these.
The thing is that metadata is not aimed to be readable for our readers, it's purely for SEO ranking purposes. Not sure how having incomplete sentences can impact the results though - so probably best to fix these.

> 
> We can always polish these manually after the fact, but I wonder what's the benefit of using an automated script if we have to manually fix it afterwards?

The purpose is not to "fix" these, but rather have the pages rank appropriately in search engines. So it's a basic description engines can leverage.

zayatamal:
> The thing is that metadata is not aimed to be readable for our readers, it's purely for SEO ranking purposes. Not sure how having incomplete sentences can impact the results though - so probably best to fix these.
> 
> >

To my understanding, incomplete sentences could lead search engines to ignore the description we added altogether and generate their own or result in lower search engine rankings (not sure about this one).

zayatamal:
> The purpose is not to "fix" these, but rather have the pages rank appropriately in search engines. So it's a basic description engines can leverage.

Regarding the automated scripts,
there is the factor that they might produce descriptions that misrepresent the intent of the page or look too vague/filler/generic maybe. (and in this case incomplete ones)

Metadata should reflect the documentation, not approximate it. If it sounds off for us, that should tell us something.

Given that, I think it makes sense to start small, with the landing pages for example. That way we get the highest impact with a manageable scope, and we can refine the approach before scaling it across the rest of the documentation.

bboisseau:
> > The thing is that metadata is not aimed to be readable for our readers, it's purely for SEO ranking purposes. Not sure how having incomplete sentences can impact the results though - so probably best to fix these.
> > >
> 
> To my understanding, incomplete sentences could lead search engines to ignore the description we added altogether and generate their own or result in lower search engine rankings (not sure about this one).

I've fixed the incomplete sentences in the most recent commit. Did you find any remaining ones?

bboisseau:
> > The purpose is not to "fix" these, but rather have the pages rank appropriately in search engines. So it's a basic description engines can leverage.
> 
> Regarding the automated scripts, there is the factor that they might produce descriptions that misrepresent the intent of the page or look too vague/filler/generic maybe. (and in this case incomplete ones)
> 
> Metadata should reflect the documentation, not approximate it. If it sounds off for us, that should tell us something.
> 
> Given that, I think it makes sense to start small, with the landing pages for example. That way we get the highest impact with a manageable scope, and we can refine the approach before scaling it across the rest of the documentation.

I think we agree on the vagueness of it, although I haven't found metadata that didn't correctly capture the purpose of said page. Happy to double check with you and others.
As for the rest, I felt like it would be a good first step to "catch-up" and have a v0 that is existing, rather than doing it incrementally on small batches. So: the v0 would have this first existing metadata, and we can incrementally improve it. (again would not be read/readable for users).

But happy to take a different route if we prefer.

PR #645, opened 2026-06-10
Document Virtualized Android/Cuttlefish

ajanon:
# Documentation changes

Document Virtualized Android/Cuttlefish for Anbox 1.30.0, and add a missing state.

# Review and preview

In your local repository,

## Reviewers

# JIRA / Launchpad bug

### DIFF
--- a/explanation/aaos.md
+++ b/explanation/aaos.md
@@ -7,7 +7,7 @@ myst:
 (exp-aaos)=
 # Android Automotive OS(AAOS)
 
-Since the 1.21.0 release, in addition to providing images based on the Android Open Source Project (AOSP), Anbox Cloud also offers images based on the [Android Automotive OS (AAOS)](https://source.android.com/docs/automotive/start/what_automotive). These images help run Android in automotive infotainment systems at scale.
+Anbox Cloud offers images based on the [Android Automotive OS (AAOS)](https://source.android.com/docs/automotive/start/what_automotive). These images help run Android in automotive infotainment systems at scale.
 
 ```{important}
 - The Anbox Cloud AAOS images are based on Android Automotive and *not* Android Auto. For understanding the differences between the two, see upstream documentation on [Android Automotive and Android Auto](https://source.android.com/docs/automotive/start/what_automotive#android-automotive-android-auto).
@@ -19,7 +19,9 @@ Until the 1.24.0 release, if you created your own VHAL implementation by followi
 
 However, if a third-party VHAL implementation is loaded during Android runtime, on the Anbox Cloud Dashboard, some vehicle property values may still not be accessible while others may not be editable on the Anbox Cloud dashboard, due to [permission controls](https://source.android.com/docs/automotive/vhal/property-configuration) that categorize vehicle properties as read-only, write-only, or read-write.
 
-Anbox Cloud offers an [Anbox-specific VHAL HIDL interface](https://github.com/canonical/vendor_canonical_interfaces). By using this HIDL, the VHAL implementation allows the Anbox VHAL adapter to modify non-writable and access non-readable vehicle property values without encountering permission issues. See the {ref}`how-to guide <howto-integrate-hidl>` on instructions for setting this up with your VHAL implementation.
+Anbox Cloud offers an [Anbox-specific VHAL HIDL interface](https://github.com/canonical/vendor_canonical_interfaces) for containerized images. By using this HIDL, the VHAL implementation allows the Anbox VHAL adapter to modify non-writable and access non-readable vehicle property values without encountering permission issues. See the {ref}`how-to guide <howto-integrate-hidl>` on instructions for setting this up with your VHAL implementation.
+
+Since the 1.30.0 release, AAOS 16 is available through images with virtualized Android (`resolute:aaos16-cf:amd64` and `resolute:aaos16-cf:arm64`). These images provide native gRPC support for the vehicle HAL. See {ref}`exp-android-execution-environments` for details on the virtualized Android execution environment.
 
 ## Related topics
 
--- a/explanation/anbox-cloud.md
+++ b/explanation/anbox-cloud.md
@@ -42,7 +42,7 @@ See the following table for a comparison of features for the different variants:
 |---------|-----------------------|-------------|
 | {ref}`exp-application-streaming` | ✓ | ✓ |
 | {ref}`exp-web-dashboard` | ✓ | ✓ |
-| {ref}`Android versions <ref-provided-images>` | 14, 15 | 14, 15 |
+| {ref}`Android versions <ref-provided-images>` | 14, 15, 16 | 14, 15, 16 |
 | [Security updates](https://ubuntu.com/support) | ✓ | ✓ |
 | [Community support](https://discourse.ubuntu.com/c/project/anbox-user/148) | ✓ | ✓ |
 | [Vendor support available](https://ubuntu.com/support) | ✓* | ✓ |
@@ -52,6 +52,10 @@ See the following table for a comparison of features for the different variants:
 *\* When purchasing the Anbox Cloud Appliance through the AWS Marketplace, the Ubuntu Pro subscription does not include vendor support.*
 
 We recommend starting with the Anbox Cloud Appliance. You can choose to expand to a charmed Anbox Cloud installation later.
+
+## Android execution environments
+
+Anbox Cloud supports two ways of running Android: containerized Android, where the Android system runs directly in the LXD container, and virtualized Android, where Android runs inside a virtual machine within the LXD instance. The execution environment is determined by the image that an instance is based on. See {ref}`exp-android-execution-environments` for details on the two environments and {ref}`ref-feature-support-by-image-type` for a comparison of their supported features.
 
 ## Anbox Cloud Components
 
--- a/explanation/android-execution-models.md
+++ b/explanation/android-execution-models.md
@@ -1,0 +1,49 @@
+(exp-android-execution-environments)=
+# Android execution environments
+
+Anbox Cloud supports two ways of running Android inside an instance: **containerized Android** and **virtualized Android**. The execution environment is determined by the image that an instance is based on - you choose an image when creating an instance, and the image determines how Android runs.
+
+Both execution environments provide access to Android through the same streaming infrastructure and can coexist in the same Anbox Cloud deployment.
+
+## Containerized Android
+
+With containerized Android, the Android system runs directly inside the LXD container. This is the execution environment used by `jammy:*` images (for example, `jammy:android14:amd64`).
+
+Containerized Android supports the full set of Anbox Cloud features:
+
+- {ref}`Applications <exp-applications>` and the application lifecycle (bootstrap, updates, versions)
+- {ref}`Addons <exp-addons>` for image customisation
+- {ref}`Platform plugins <exp-platforms>` for custom rendering and input pipelines
+- Shell access to the Android environment through `anbox-shell`
+
+This is the execution environment that Anbox Cloud has used since its first release. The Android system shares the kernel with the host through LXD's container isolation, which keeps resource overhead low and allows high instance density.
+
+## Virtualized Android
+
+With virtualized Android, the Android system runs inside a [Cuttlefish](https://source.android.com/docs/devices/cuttlefish) virtual machine within the LXD instance. This is the execution environment used by `resolute:*-cf:*` images (for example, `resolute:android16-cf:amd64`). The `-cf` suffix in the image name indicates that the image uses the Cuttlefish virtual device.
+
+Cuttlefish is Google's reference virtual Android device. Running Android through Cuttlefish means you get a standard, unmodified Android environment with no Anbox-specific changes to the Android system itself — the Android system image comes directly from Google's build infrastructure. This is the right choice when you need Android to behave exactly as it does on a physical device or in Google's own test environments.
+
+Virtualized Android supports {ref}`instances <sec-application-raw-instances>`. To access the Android shell, use `adb shell` instead of `anbox-shell`. See {ref}`ref-feature-support-by-image-type` for the full feature comparison.
+
+Virtualized Android is a good fit for the following scenarios:
+
+- **Standard Android environments** where you need Android to behave exactly as it does on a real device, without any Anbox-specific modifications to the Android system.
+- **Custom Android or AAOS builds** where you want to run your own Android system image inside Anbox Cloud. See {ref}`howto-package-custom-android-build` for instructions.
+- **VHAL development** where native gRPC support for the vehicle HAL simplifies automotive development.
+- **Workloads that benefit from stronger isolation** where the additional virtualisation boundary between Android and the host is desirable.
+
+## Choosing between the two environments
+
+Use **virtualized Android** when you need a standard Android environment that behaves exactly like a physical device or Google's reference implementation, when you want to run a custom Android or AAOS build, or when you need native VHAL support for automotive use cases.
+
+Use **containerized Android** when you are building on top of Anbox Cloud's application and addon model for managed APK deployment, or when you need platform plugins for custom rendering and input pipelines.
+
+Both image types can coexist in the same AMS deployment. You can register and use images of both types simultaneously and launch instances from either type as needed.
+
+## Related topics
+
+- {ref}`ref-feature-support-by-image-type`
+- {ref}`tut-getting-started-virtualized-android`
+- {ref}`howto-package-custom-android-build`
+- {ref}`ref-provided-images`
--- a/explanation/images.md
+++ b/explanation/images.md
@@ -18,6 +18,7 @@ The following table lists the different statuses that an image can have dependin
 | `initializing` | The image is currently being created. |
 | `created` | The image was uploaded to AMS but not yet available on all LXD nodes. |
 | `available` | The image is present on the remote but not in the LXD cluster. |
+| `queued` | The image operation is queued for a download. |
 | `active` | The image is available on all LXD nodes. |
 | `deleted` | The image has been deleted and no longer available for use. |
 | `error` | The image has encountered an error. |
--- a/explanation/images.md
+++ b/explanation/images.md
@@ -25,3 +25,12 @@ The following table lists the different statuses that an image can have dependin
 | `unknown` | A possible error occurred and the real state of the image cannot be determined. |
 
 If you encounter the `error` or the `unknown` status, [file a bug](https://bugs.launchpad.net/anbox-cloud).
+
+## Image variants
+
+Anbox Cloud provides two variants of images that use different Android execution environments:
+
+- **Images with containerized Android** (`jammy:*`) run Android directly in the LXD container and support the full application model, addons, and platform plugins.
+- **Images with virtualized Android** (`resolute:*-cf:*`) run Android inside a virtual machine within the LXD instance.
+
+The image variant determines which execution environment is used. See {ref}`exp-android-execution-environments` for an explanation of the two environments and {ref}`ref-feature-support-by-image-type` for a detailed feature comparison.
--- a/explanation/security/anbox-runtime.md
+++ b/explanation/security/anbox-runtime.md
@@ -31,8 +31,18 @@ For streaming of audio, video and other data, the Anbox runtime uses WebRTC as p
 
 The security model and cryptographic use of WebRTC is described in [RFC8827](https://www.rfc-editor.org/rfc/rfc8827) and use of WebRTC in the Anbox runtime does not deviate from this.
 
-## Packages used
+## Packages used for containerized Android
 
 - [Go standard library](https://pkg.go.dev/std)
 - [OpenSSL](https://launchpad.net/ubuntu/+source/openssl/)
 - [`ca-certificates`](https://launchpad.net/ubuntu/+source/ca-certificates)
+
+## Virtualized Android
+
+The runtime for virtualized Android is written in Rust and runs within snap confinement. Its security model adds the following properties on top of the general security measures described above:
+
+* **Rust memory safety** - the Rust ownership model eliminates common memory safety vulnerabilities (buffer overflows, use-after-free, data races) at compile time.
+* **Snap confinement** - the runtime runs as a snap with strict confinement, limiting its access to the host system.
+* **QEMU isolation** - Android runs inside a virtual machine, providing an additional isolation boundary between the Android workload and the host.
+
+See {ref}`exp-android-execution-environments` for more details on the two execution environments.
--- a/howto/images/package-custom-android-build.md
+++ b/howto/images/package-custom-android-build.md
@@ -1,0 +1,160 @@
+---
+myst:
+  html_meta:
+    "description": "How to package a custom Android Cuttlefish build."
+---
+
+(howto-package-custom-android-build)=
+# Package a custom Android build
+
+You can package a custom [Cuttlefish](https://source.android.com/docs/devices/cuttlefish) Android build as a [snap](https://snapcraft.io/) and install it in a running virtualized Android instance. This lets you run your own Android system image while using the Anbox Cloud infrastructure for streaming and instance management.
+
+For background on how virtualized Android works in Anbox Cloud, see {ref}`exp-android-execution-environments`.
+
+## Prerequisites
+
+- An Anbox Cloud deployment with a virtualized Android image available (see {ref}`tut-getting-started-virtualized-android`)
+- Android Cuttlefish build artifacts from your own build:
+  * `cvd-host_package.tar.gz` - Cuttlefish host tools from the Android build
+  * Android images (such as `super.img`, `boot.img`, `vbmeta.img`, and other system images produced by the build)
+- The `snap` command-line tool installed on your build machine
+
+## Prepare the snap directory structure
+
+Create the directory structure for the snap:
+
+```bash
+mkdir -p my-android-snap/host-tools
+mkdir -p my-android-snap/images
+mkdir -p my-android-snap/meta
+```
+
+Extract the Cuttlefish host tools:
+
+    tar xf cvd-host_package.tar.gz -C my-android-snap/host-tools/
+
+Copy the Android system images:
+
+    cp super.img boot.img vbmeta.img vendor_boot.img android-info.txt \
+        my-android-snap/images/
+
+Include all image files that your Cuttlefish build produced and that are required to launch a virtual device.
+
+## Create the snap metadata
+
+Create the file `my-android-snap/meta/snap.yaml`:
+
+```yaml
+name: my-custom-android
+version: '1.0'
+summary: Custom Android Cuttlefish images and host tools
+description: |
+  Provides custom Android system images and Cuttlefish host tools
+  for use with the Anbox Cloud runtime.
+type: app
+base: bare
+architectures:
+  - amd64
+confinement: strict
+grade: stable
+slots:
+  host-tools:
+    interface: content
+    content: android-host-tools
+    read:
+      - $SNAP/host-tools
+  images:
+    interface: content
+    content: android-images
+    read:
+      - $SNAP/images
+```
+
+The snap defines two content interface slots that provide the Cuttlefish host tools and Android system images to the Anbox runtime snap. If you are targeting ARM64, change the `architectures` field to `arm64`.
+
+## Pack the snap
+
+Use `snap pack` to create the snap:
+
+    snap pack my-android-snap/
+
+This produces a file like `my-custom-android_1.0_amd64.snap` in the current directory.
+
+## Launch an instance and install the snap
+
+Launch a virtualized Android instance in development mode so that you can push files to it:
+
+    id="$(amc launch --devmode resolute:android16-cf:amd64)"
+
+Or, for an automotive Android build:
+
+    id="$(amc launch --devmode resolute:aaos16-cf:amd64)"
+
+Wait for the instance to reach the `running` state:
+
+    amc wait "${id}" --timeout 15m
+
+See {ref}`sec-dev-mode` for more information about development mode.
+
+Push the snap into the instance:
+
+    cat my-custom-android_1.0_amd64.snap \
+      | amc exec "${id}" -- sh -c "cat - > /tmp/my-custom-android_1.0_amd64.snap"
+
+Install the snap inside the instance:
+
+    amc exec "${id}" -- snap install --dangerous /tmp/my-custom-android_1.0_amd64.snap
+
+The `--dangerous` flag is required because the snap is not signed by the Snap Store.
+
+## Connect snap interfaces and restart the runtime
+
+Disconnect the existing Android snap from the runtime:
+
+    amc exec "${id}" -- snap disconnect anbox-runtime:android-images
+    amc exec "${id}" -- snap disconnect anbox-runtime:android-host-tools
+
+Connect your custom Android snap to the runtime:
+
+    amc exec "${id}" -- snap connect anbox-runtime:android-images my-custom-android:images
+    amc exec "${id}" -- snap connect anbox-runtime:android-host-tools my-custom-android:host-tools
+
+Restart the runtime to load the new Android images:
+
+    amc exec "${id}" -- snap restart anbox-runtime
+
+## Verify
+
+Connect to the instance and verify that your custom Android build is running:
+
+    amc exec "${id}" -- adb shell getprop ro.build.display.id
+
+The output should show the build ID of your custom Android build.
+
+## Troubleshooting
+
+### Snap content interface connection fails
+
+If `snap connect` fails, verify that the content interface names in your snap match what the runtime expects:
+- The `content` field for host tools must be `android-host-tools`
+- The `content` field for images must be `android-images`
+
+Use `snap connections anbox-runtime` inside the instance to inspect the expected interfaces.
+
+### Runtime fails to start after connecting custom snap
+
+Check the runtime logs:
+
+    amc exec "${id}" -- snap logs anbox-runtime -n 100
+
+Common issues include missing image files or incompatible Cuttlefish host tool versions.
+
+### Snap installation fails
+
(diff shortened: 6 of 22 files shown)


reference/feature-support-by-image-type.md
```diff
@@ -0,0 +1,58 @@
...
+
+- {ref}`Applications <exp-applications>` — managed APK deployment with versioning, bootstrapping, and the full application lifecycle
+- {ref}`Addons <exp-addons>` — customising images with hooks tied to instance lifecycle events
+- {ref}`Platform plugins <exp-platforms>` — custom rendering and input pipelines via the Anbox Platform SDK
+- Streaming
+- Custom Android builds (see {ref}`exp-custom-images`)
+- Shell access via `anbox-shell`
+- VHAL support via adapter
+
+**Minimum resource requirements:** 3 GB memory per instance. KVM not required.
```
adglkh: Maybe we should also mention 2 CPU cores and 3 GiB of disk space here as the minimum requirements?

reference/feature-support-by-image-type.md
```diff
@@ -0,0 +1,58 @@
...
+
+The Android system runs inside a Cuttlefish virtual machine within the LXD instance, providing a standard, unmodified Android environment. These images support:
+
+- Streaming
+- Custom Android or AAOS builds (see {ref}`howto-package-custom-android-build`)
+- Shell access via `adb shell` on port 6520
+- VHAL support with native gRPC
+- QEMU-based isolation between Android and the host
+
+**Minimum resource requirements:** 5 GB memory per instance. KVM required on the host.
```
adglkh: Same above.

For cuttlefish, the minimal requirement is 4 CPU cores, 5 GB memory and 15 GiB of disk space

howto/images/package-custom-android-build.md
```diff
@@ -0,0 +1,147 @@
...
+
+    snap pack my-android-snap/
+
+This produces a file like `my-custom-android_1.0_amd64.snap` in the current directory.
+
+## Launch an instance and install the snap
+
+Launch a virtualized Android instance in development mode so that you can push files to it:
+
+    id=$(amc launch --devmode --raw resolute:android16-cf:amd64 --memory 5GB)
```
adglkh: We could drop `--memory 5GB` here, as AMS will automatically assign 5GB of memory to the instance by default for Cuttlefish image.

howto/images/package-custom-android-build.md
```diff
@@ -0,0 +1,147 @@
...
+
+Wait for the instance to reach the `running` state:
+
+    amc wait "${id}" --timeout 5m
+
+See {ref}`sec-dev-mode` for more information about development mode.
+
+Push the snap into the instance:
+
+    lxc file push my-custom-android_1.0_amd64.snap "ams-${id}/tmp/"
```
adglkh: For the appliance scenario, we need to append `--project=anbox-cloud` to the above command.
For the charm based deployment, running the above command require people to goto ams-lxd unit first, so command like
`juju ssh lxd/leader` is required. 

For the sake of convenience and simplicity, I'd suggest using
```
cat  my-custom-android_1.0_amd64.snap | amc exec "${id}" -- sh -c "cat - > /tmp/my-custom-android_1.0_amd64.snap"
``` 
which should work for both cases.

explanation/android-execution-models.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Virtualized Android
+
+With virtualized Android, the Android system runs inside a [Cuttlefish](https://source.android.com/docs/devices/cuttlefish) virtual machine within the LXD instance. This is the execution model used by `resolute:*-cf:*` images (for example, `resolute:android16-cf:amd64`). The `-cf` suffix in the image name indicates that the image uses the Cuttlefish virtual device.
+
+Cuttlefish is Google's reference virtual Android device. Running Android through Cuttlefish means you get a standard, unmodified Android environment with no Anbox-specific changes to the Android system itself — the Android system image comes directly from Google's build infrastructure. This is the right choice when you need Android to behave exactly as it does on a physical device or in Google's own test environments.
+
+Virtualized Android supports {ref}`instances <sec-application-raw-instances>`. To access the Android shell, use `adb shell` on port 6520 instead of `anbox-shell`. See {ref}`ref-feature-support-by-image-type` for the full feature comparison.
```
adglkh: > To access the Android shell, use `adb shell` on port 6520 instead of `anbox-shell`

A: this line is a bit confusing. I think the author wanted to explain that to access the Android shell from the outside, you should use `adb shell <instance_id>:6520`. However, `anbox-shell` is meant to be run inside the instance. Since we pre-install adb within the cuttlefish image anyway, we could rephrase that line to:

"To access the Android shell, use `adb shell` instead of `anbox-shell`."

explanation/android-execution-models.md
```diff
@@ -0,0 +1,57 @@
...
+
+Virtualized Android is a good fit for the following scenarios:
+
+- **Standard Android environments** where you need Android to behave exactly as it does on a real device, without any Anbox-specific modifications to the Android system.
+- **Custom Android or AAOS builds** where you want to run your own Android system image inside Anbox Cloud. See {ref}`howto-package-custom-android-build` for instructions.
+- **VHAL development** where native gRPC support for the vehicle HAL simplifies automotive development.
+- **Workloads that benefit from stronger isolation** where the additional virtualisation boundary between Android and the host is desirable.
+
+Virtualized Android instances require a minimum of 5 GB of memory and KVM support on the host.
```
adglkh: See below, do we want to mention the minimal requirement for cpu and disk-size here as well?

reference/feature-support-by-image-type.md
```diff
@@ -0,0 +1,58 @@
...
+
+**Minimum resource requirements:** 3 GB memory per instance. KVM not required.
+
+## Images with virtualized Android (`resolute:*-cf:*`)
+
+The Android system runs inside a Cuttlefish virtual machine within the LXD instance, providing a standard, unmodified Android environment. These images support:
+
+- Streaming
+- Custom Android or AAOS builds (see {ref}`howto-package-custom-android-build`)
+- Shell access via `adb shell` on port 6520
```
zayatamal: should this be added to network-ports.md?

reference/feature-support-by-image-type.md
ajanon: Wasn't it 10GiB for CF? (from the top of my head)

howto/images/package-custom-android-build.md
ajanon: That's kinda hacky, but you're right.

explanation/android-execution-models.md
ajanon: This would duplicate information here and at reference/feature-support-by-image-type

I'm fine with that if we think it's better.

@zayatamal Any opinion?

explanation/images.md
```diff
@@ -22,5 +22,15 @@ The following table lists the different statuses that an image can have dependin
 | `deleted` | The image has been deleted and no longer available for use. |
 | `error` | The image has encountered an error. |
 | `unknown` | A possible error occurred and the real state of the image cannot be determined. |
+| `queued` | The image operation is queued and waiting to be processed. |
```
zayatamal: a minor detail but `queued` status is appended at the end of the table but maybe should come before `initializing` (correct me if i am wrong, an image is queued before it starts initializing)

reference/feature-support-by-image-type.md
ajanon: Hm...  We don't mention the regular adb port which is at 5559 for regular instances. (though we prefer people using amc connect anyway IIRC)

@adglkh any opinion?

explanation/images.md
ajanon: I would need to check in the code tbh. (or maybe @jat-canonical could chime in).

Do you think we should have an order for these? Because arguably, deleted, error, and unknown are not really in any order.

I'm fine with changing if you prefer anyway.

tutorial/getting-started-with-virtualized-android.md
zayatamal, line 1: I'll add some comments but would need your input with the technical details.

- This section in its current state reads like a how to guide in a way. and does not read like a complete arc.
It states the end goal "At the end, you will have a running Android instance that you can interact with through ADB." but ends at the clean up, then what. So the complete arc of the tutorial is: Launch instance → open a shell → delete instance?
the "aha moment" a tutorial should deliver is: you are streaming Android 16 (based on information from explanation) but I do not reach it here.

- tutorials use "we" so the opening would be:  In this tutorial, we will launch an instance using a virtualized Android image and connect to it. By the end of this tutorial, we will have a running Android instance that we can interact with through ADB and [stream through the Anbox Cloud dashboard]<< from my understanding reading the other segments we can stream it.

- change all the "yous" to "we"

- we also jump from amc shell test0 then adb shell (no connect step) whereas in another page i think it was access instance:  adb connect <ip>:6520 then adb shell (this step is important in a tutorial)

- add a success segment: ex: We have successfully launched a virtualized Android instance. 

- add this line in: Prerequisites
To proceed with the tutorial, we need:
-"Next:" pointer to the "how-to-guides"

tutorial/index.md
zayatamal, line 1: this feels a bit off, maybe:
The tutorials in this section guide you through the workflows of {term}`Anbox Cloud Appliance`, from installation to running and streaming Android instances.

The following tutorials use images with containerized Android.

- {ref}`tut-installing-appliance` 
- {ref}`tut-create-virtual-device` 
- {ref}`tut-set-up-stream-client` 

The following tutorial uses images with virtualized Android.

- {ref}`tut-getting-started-virtualized-android` 

For an explanation of the two execution models, see {ref}exp-android-execution-models.

reference/anbox-features.md
```diff
@@ -11,6 +11,8 @@ Anbox Cloud provides images based on the Android Open Source Project (AOSP), an
 
 The following table lists some Anbox features and whether they are supported for a given base.
 
+The following feature table applies to images with containerized Android only. For a comparison of feature support between containerized and virtualized Android, see {ref}`ref-feature-support-by-image-type`.
```
zayatamal: very minor and subtle change to leave no room for ambiguity:
the word "only" should precede what it modifies. so sentence would be : The following feature table applies only to images with containerized Android.

reference/feature-support-by-image-type.md
zayatamal: I wanted to check since I noticed some mentioned

explanation/images.md
zayatamal: If there is no reason for reordering then i am fine by it.

explanation/android-execution-models.md
zayatamal: they are already mentioned in reference/feature-support and in the tutorial. 
Other than that, and abiding by the framework, the actual numbers belong in reference. if we need to do anything here, in the explanation page, it would probably be answering why we need these requirements

explanation/aaos.md
```diff
@@ -9,6 +9,8 @@ myst:
 
 Since the 1.21.0 release, in addition to providing images based on the Android Open Source Project (AOSP), Anbox Cloud also offers images based on the [Android Automotive OS (AAOS)](https://source.android.com/docs/automotive/start/what_automotive). These images help run Android in automotive infotainment systems at scale.
 
+Since 1.30, AAOS 16 is available through images with virtualized Android (`resolute:aaos16-cf:amd64` and `resolute:aaos16-cf:arm64`). These images provide native gRPC support for the vehicle HAL. See {ref}`exp-android-execution-models` for details on the virtualized Android execution model.
```
zayatamal: minor but Since the 1.30.0 release 
and another note, shouldn't the {important} admonishment below come before any version-specific detail (just an observation)

also this page in the documentation goes on to explain the VHAL adapter introduced in 1.24.0 for containerized AAOS images . I noticed that we don't have context for "native gRPC" (virtualized) as is done for the former.

reference/feature-support-by-image-type.md
adglkh: https://github.com/canonical/ams/blob/main/pkg/resourcepreset/resourcepreset.go#L27

It's 15 GiB for CF

reference/feature-support-by-image-type.md
adglkh: > network-ports.md

A: The network-ports.md page lists all network ports required at the core service level, not the instance level. I don't think so.

adglkh: > Hm... We don't mention the regular adb port which is at 5559 for regular instances. (though we prefer people using amc connect anyway IIRC)
> 
> @adglkh any opinion?

A: usually the common pracitse for people to use the adb are two ways:
    
    a). amc launch -r <instance_id> -s adb
        then
        adb connect <internal_ip_addr_instance>
    or
    b). amc connect after sharing adb connection via `anbox-stream-gateway session share`

I am not sure we suggest people directly using `adb connect <instance_id>:6520` here.

@morphis could you share you thoughts ?

explanation/images.md
jat-canonical: I agree with Amal here that the order should semantically make sense but the status should come after `available` as the `queued` status, in its current state, is set only when a download has been triggered for the image. So in addition we should rather specify that `the image is queued for a download` for the description than putting something vague like `waiting to be processed` .

explanation/android-execution-models.md
ajanon: I'm removing the line from now, but we can think of explaining if we feel the need.

explanation/images.md
ajanon: I changed it.

tutorial/getting-started-with-virtualized-android.md
ajanon: Okay, makes sense. I think we need to discuss this a bit more as that would be a rather large change. (or feel free to make a commit as we discussed).

tutorial/index.md
ajanon: Can I take this as-is?

explanation/aaos.md
ajanon: > shouldn't the {important} admonishment below come before any version-specific detail (just an observation)

Sure, do you think we should make the change right away?

> I noticed that we don't have context for "native gRPC" (virtualized) as is done for the former.

If somebody brings their own implementation which talks using gRPC, it should work out of the box.

ajanon: We need to rewrite this somewhat, as the HIDL paragraph further below is only valid for container images.

tutorial/index.md
zayatamal: yes

howto/images/package-custom-android-build.md
```diff
@@ -0,0 +1,148 @@
...
+Copy the Android system images:
+
+    cp super.img boot.img vbmeta.img vendor_boot.img android-info.txt \
+       my-android-snap/images/
+
+Include all image files that your Cuttlefish build produced and that are required to launch a virtual device.
+
+## Create the snap metadata
+
+Create the file `my-android-snap/meta/snap.yaml`:
```
zayatamal, line 45: the [code block ](https://documentation.ubuntu.com/sphinx-stack/latest/reference/myst-syntax/#code-blocks)below should follow yaml code block, i linked it above:

````
```yaml
# Demonstrate a code block
code:
- example: true
```
````

File: howto/images/package-custom-android-build.md [hunk 12]
```diff
@@ -0,0 +1,148 @@
...
+    cat my-custom-android_1.0_amd64.snap \
+    | amc exec "${id}" -- sh -c "cat - > /tmp/my-custom-android_1.0_amd64.snap"
+
+Install the snap inside the instance:
+
+    amc exec "${id}" -- snap install --dangerous /tmp/my-custom-android_1.0_amd64.snap
+
+The `--dangerous` flag is required because the snap is not signed by the Snap Store.
+
+## Connect snap interfaces and restart the runtime
```
zayatamal, line 110: I feel like this section "Connect snap interfaces and restart the runtime" has all actions under one sentence and then disconnect/connect/restart block is one large block. so to ease it off, it could be written like this

Disconnect the existing Android snap from the runtime:
   amc exec "${id}" -- snap disconnect anbox-runtime:android-images
   amc exec "${id}" -- snap disconnect anbox-runtime:android-host-tools

Connect your custom Android snap to the runtime:
    amc exec "${id}" -- snap connect anbox-runtime:android-images my-custom-android:images
    amc exec "${id}" -- snap connect anbox-runtime:android-host-tools my-custom-android:host-tools

Restart the runtime to load(pick up depending on which verb describes it best) the new Android images:
 amc exec "${id}" -- snap restart anbox-runtime

howto/images/package-custom-android-build.md
```diff
@@ -0,0 +1,148 @@
...
+
+    snap pack my-android-snap/
+
+This produces a file like `my-custom-android_1.0_amd64.snap` in the current directory.
+
+## Launch an instance and install the snap
+
+Launch a virtualized Android instance in development mode so that you can push files to it:
+
+    id="$(amc launch --devmode resolute:android16-cf:amd64)"
```
zayatamal, line 87: quick question here, should we add  somewhere in this how to guide page that: this guide applies to both AOSP and AAOS custom builds. Replace `resolute:android16-cf:amd64` with `resolute:aaos16-cf:amd64` (or the `arm64` variants) when working with a custom AAOS build.

the reason i am asking is because there are at least 3 pages (feature support in reference, android execution models in explanation, and aaos.md) referring to this how-to page when they mention AAOS but it is not acknowledged. 
which in theory(and without being explicitly mentioned) they could replace resolute:android16-cf:amd64 with  resolute:aaos16-cf:amd64 if I understand correctly

howto/instance/access-instance.md
```diff
@@ -20,6 +20,15 @@ This command opens a bash shell inside the instance. To access the nested Androi
 
     amc exec <id> -- anbox-shell
 
+```{note}
```
zayatamal: i think this note feels bolted on and not integrated with the flow here. and it's causing a bit of an issue because if I go over the page:
I follow the main text, run amc shell <id>, gets a bash shell
read "use the anbox-shell command inside the new shell"
run anbox-shell but would not work if it's a virtualized image
eventually find the note at the bottom. (It would be a quick patch for now)
so i thought why not showcase both as options instead of making the other seem like an afterthought.

would this alternative be correct?:
# Access an instance

In some cases, it might be necessary to access an individual instance for debugging reasons. You can do this on the command line with the amc command.

## Access an instance with amc

The amc command provides simple shell access to any instance managed by AMS. To access a specific instance you only need its ID:

amc shell <id>

This command opens a bash shell inside the instance.

## Access the Android environment (couldn't think of a different heading)


### Containerized Android (`jammy:*` images)

Use anbox-shell to access the Android container. If you combine the anbox-shell command with amc exec, you can get direct access to the Android container:

amc exec <id> -- anbox-shell

If you only want to watch the Android log output, use the following command:

amc exec <id> -- anbox-shell logcat

amc shell and amc exec open various possibilities for automation use cases. See the help output of the commands for further details.

### Virtualized Android (`resolute:*-cf:*` images)

Use ADB instead of anbox-shell:

adb connect <instance_ip>:6520
adb shell

explanation/android-execution-models.md
```diff
@@ -0,0 +1,55 @@
...
+
+Use **containerized Android** when you are building on top of Anbox Cloud's application and addon model for managed APK deployment, or when you need platform plugins for custom rendering and input pipelines.
+
+Both image types can coexist in the same AMS deployment. You can register and use images of both types simultaneously and launch instances from either type as needed.
+
+## Runtime convergence
+
+The two execution models currently use different runtime implementations. This is an implementation detail — both models present the same user-facing capabilities through AMS and the streaming stack.
+
+The runtime that currently powers virtualized Android will also support containerized Android in a future release, converging the two implementations into a single runtime. The two execution models — containerized and virtualized — will remain as distinct choices for users.
```
zayatamal: this is another minor detail, but this describes something that has not happened yet taking into account that documentation describes the current state of the product. this is reading like a roadmap statement which belongs in release announcements, not in explanation pages. 

We should either remove it or mark it as a planned future direction with an  admonishment.

explanation/security/anbox-runtime.md
```diff
@@ -33,6 +33,18 @@ The security model and cryptographic use of WebRTC is described in [RFC8827](htt
 
 ## Packages used
```
zayatamal: unless i am mistaken (please correct me if I am wrong) , this heading (packages used) does not properly or accurately describe the pacakges used with virtualized Android part but I cannot think of a more suitable heading

reference/requirements.md
```diff
@@ -57,6 +57,8 @@ The following hardware specifications are required for running the appliance:
 
 The above recommendation is the minimum requirement to run the appliance. As Anbox Cloud is dependent on available resources to launch its Android containers, the available resources dictate the maximum number of possible Android containers. See {ref}`exp-capacity-planning` for an explanation on how to plan for a specific capacity on your appliance.
 
+Images with virtualized Android (`resolute:*-cf:*`) have additional requirements: the host CPU must support hardware virtualisation and `/dev/kvm` must be available, and each instance requires a minimum of 5 GB of memory. See {ref}`ref-feature-support-by-image-type` for details.
```
zayatamal: since this is the **requirements page**, we should mention them here: 4 CPU cores, Disk: 15 GiB 

i know that there is a reference to feature comparison page, but this is the requirement page, they should be here

tutorial/getting-started-with-virtualized-android.md
```diff
@@ -0,0 +1,51 @@
...
+
+      ls /dev/kvm
+
+## Launch an instance
+
+Launch an instance from the image. Virtualized Android instances require at least 4 CPU cores, 5 GB of memory and 15 GB of disk space:
+
+    amc launch resolute:android16-cf:amd64 --cpus 4 --memory 5GB --disk-size 15GB --name test0
+
+Note the instance ID in the output. Wait for the instance to reach the `running` state:
```
zayatamal, line 29: what happened to this?

explanation/security/anbox-runtime.md
adglkh: I feel the same way. In the case of the virtualized Android, it is more about the technical design choices and the programming languages involved.

explanation/anbox-cloud.md
```diff
@@ -53,6 +53,10 @@ See the following table for a comparison of features for the different variants:
 
 We recommend starting with the Anbox Cloud Appliance. You can choose to expand to a charmed Anbox Cloud installation later.
 
+## Android execution models
```
adglkh: Still, I am not a fan of this, as it doesn't clearly differentiate between containerized and virtualized solutions in terms of terminology.

Maybe one of the following
- Android deployment model
- Android isolate model
- Android execution environment

or do you have a better term for this?

tutorial/getting-started-with-virtualized-android.md
adglkh: For the CF instance, it usually takes longer to launch than a container based instance, especially for the very first time as it runs in VM

reference/feature-support-by-image-type.md
```diff
@@ -0,0 +1,66 @@
...
+| Pattern | Execution model | Example |
+|---------|----------------|---------|
+| `jammy:*` | Containerized Android | `jammy:android14:amd64` |
+| `resolute:*-cf:*` | Virtualized Android | `resolute:android16-cf:amd64` |
+
+The `-cf` suffix identifies a Cuttlefish-based image using virtualized Android.
+
+See {ref}`ref-provided-images` for the full list of available images.
+
+## Images with containerized Android (`jammy:*`)
```
adglkh: I'd suggest we drop `(`jammy:*`)` in the header and describe instead briefely like in the following paragraf
```
The Android system runs directly in the LXD container. They typically follow the distro:version:* pattern (for example, jammy:*).
```

Same below

explanation/anbox-cloud.md
zayatamal: my thoughts for the Android deployment model suggestion: deployment in Anbox Cloud already refers to appliance vs charmed (in my understanding) so it would collide immediately and maybe confuse users.

for the execution environment suggestion:
if I am thinking between Android execution model and Android execution environment, I will be assessing what model and environment would mean in such context. linguistically, a model describes how something operates ( which would be, in the case of Anbox Cloud the image you choose before creating an instance), while an environment describes where it operates (what is there/happening after the instance is running) . 

regarding the Android isolate model, I have no thoughts :(

your thoughts @ajanon @morphis?

howto/images/package-custom-android-build.md
adglkh: > quick question here, should we add somewhere in this how to guide page that: this guide applies to both AOSP and AAOS custom builds. Replace `resolute:android16-cf:amd64` with `resolute:aaos16-cf:amd64` (or the `arm64` variants) when working with a custom AAOS build.

A: that's a really good point. IMO yes.

> 
> the reason i am asking is because there are at least 3 pages (feature support in reference, android execution models in explanation, and aaos.md) referring to this how-to page when they mention AAOS but it is not acknowledged. which in theory(and without being explicitly mentioned) they could replace resolute:android16-cf:amd64 with resolute:aaos16-cf:amd64 if I understand correctly

A: yes, you're right.

howto/instance/access-instance.md
adglkh: +1, in my view, the above version looks better.

reference/provided-images.md
```diff
@@ -15,19 +15,25 @@ Anbox Cloud images are regular [Ubuntu cloud images](https://cloud-images.ubuntu
...
-| `jammy:aaos15:arm64`        | AAOS     | 15              | 22.04          | 1.26.0 |
-| `jammy:android15:amd64`     | AOSP     | 15              | 22.04          | 1.26.0 |
-| `jammy:android15:arm64`     | AOSP     | 15              | 22.04          | 1.26.0 |
-| `jammy:aaos14:amd64`        | AAOS     | 14              | 22.04          | 1.24.0 |
-| `jammy:aaos14:arm64`        | AAOS     | 14              | 22.04          | 1.24.0 |
-| `jammy:android14:amd64`     | AOSP     | 14              | 22.04          | 1.24.0 |
-| `jammy:android14:arm64`     | AOSP     | 14              | 22.04          | 1.24.0 |
+| Name                        | Based on | Android Version | Ubuntu Version | Execution model | Available since |
+|-----------------------------|----------|-----------------|----------------|-----------------|---------------|
+| `resolute:android16-cf:amd64` | AOSP  | 16              | 24.04          | Virtualized | 1.30.0 |
```
adglkh: Ubuntu "Resolute Raccoon" -> Ubuntu 26.04 LTS, 24.04 was wrong to me.
Same for other places below.

reference/provided-images.md
```diff
@@ -15,19 +15,25 @@ Anbox Cloud images are regular [Ubuntu cloud images](https://cloud-images.ubuntu
...
+| `jammy:android15:arm64`     | AOSP     | 15              | 22.04          | Containerized | 1.26.0 |
+| `jammy:aaos14:amd64`        | AAOS     | 14              | 22.04          | Containerized | 1.24.0 |
+| `jammy:aaos14:arm64`        | AAOS     | 14              | 22.04          | Containerized | 1.24.0 |
+| `jammy:android14:amd64`     | AOSP     | 14              | 22.04          | Containerized | 1.24.0 |
+| `jammy:android14:arm64`     | AOSP     | 14              | 22.04          | Containerized | 1.24.0 |
 
 ## Support for Anbox Cloud images
 
-Currently, Anbox Cloud provides images based on Ubuntu 22.04 LTS (Jammy Jellyfish). Deprecations, if any, are announced at least two releases in advance.
+Currently, Anbox Cloud provides images based on Ubuntu 22.04 LTS (Jammy Jellyfish) and Ubuntu 24.04 LTS (Noble Numbat). Deprecations, if any, are announced at least two releases in advance.
```
adglkh: Ditto

tutorial/getting-started-with-virtualized-android.md
```diff
@@ -0,0 +1,51 @@
...
+## Prerequisites
+
+- A working Anbox Cloud deployment (appliance or charmed). If you haven't set one up yet, see {ref}`tut-installing-appliance`.
+- KVM support on the machine running Anbox Cloud. Verify by checking that `/dev/kvm` exists:
+
+      ls /dev/kvm
+
+## Launch an instance
+
+Launch an instance from the image. Virtualized Android instances require at least 4 CPU cores, 5 GB of memory and 15 GB of disk space:
```
zayatamal: add them in the prerequisites

tutorial/getting-started-with-virtualized-android.md
```diff
@@ -0,0 +1,51 @@
...
+
+    adb shell
+
+You now have a shell inside the Android system running in your instance.
+
+## View Android logs
+
+To view the Android system logs:
+
+    adb logcat
```
zayatamal: because of the last step: we can have it as logcat (make sure that it's true)

tutorial/getting-started-with-virtualized-android.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Clean up
```
zayatamal: we can remove this,
add a success part

howto/instance/access-instance.md
zayatamal: remove images from the titles, maybe reference the difference between them or add them as sentences after the heading

reference/feature-support-by-image-type.md
ajanon: I went with amc exec -- adb shell instead.

This provides consistency with anbox-shell.

We can revisit in the future, but given the short time, let's do it this way for now.

tutorial/getting-started-with-virtualized-android.md
zayatamal: should we add Gary's point as a note?  or would   amc wait test0 --timeout 15m be enough

tutorial/getting-started-with-virtualized-android.md
```diff
@@ -0,0 +1,49 @@
...
+
+Open a shell inside the `test0` instance by running
+
+    amc shell test0
+
+Connect to the Android shell through ADB
+
+    adb shell
+
+We have successfully launched a virtualized Android instance with ADB shell access.
```
zayatamal, line 45: Can we have a success heading before this one. Thank you!!

reference/glossary.md
```diff
@@ -273,6 +276,9 @@ Ubuntu One
 
     See [Ubuntu One](https://login.ubuntu.com/) for more information.
 
+Virtualized Android
+    An {ref}`Android execution model <exp-android-execution-environments>` where the Android system runs inside a Cuttlefish virtual machine within the LXD instance. Used by `resolute:*-cf:*` images.
```
zayatamal: this is causing a confusion 
model and environment

tutorial/getting-started-with-virtualized-android.md
ajanon: What about now?

tutorial/getting-started-with-virtualized-android.md
```diff
@@ -0,0 +1,51 @@
...
+
+In this tutorial, we will go through launching our first instance with virtualized Android and connecting to it. At the end, we will have a running Android instance that we can interact with through ADB.
+
+## Prerequisites
+
+To proceed with the tutorial, we need:
+- A working Anbox Cloud deployment (appliance or charmed). If you haven't set one up yet, see {ref}`tut-installing-appliance`. It needs at least:
+  * 4 CPU cores
+  * 5GB of memory
+  * 15 GB of disk space
```
zayatamal, line 18: very very minor but in some pages we have    15 GB of disk space and other pages 15GiB of disk space (like in the reference/requirements.md) @ajanon

ajanon: I actually don't know, this is always confusing and we're not internally consistent in AMS anyway IIRC. I will try to fix it.

reference/feature-support-by-image-type.md
adglkh: Sorry, I don't get response on this one yet.
@zayatamal @ajanon do we want to keep it as is or drop (`jammy:*`) as I suggested above?

zayatamal: i think we can drop it as Gary said and have it a part of the paragraph.

adglkh (commented):
Thank you both.
We're getting there. 

One last thing come into my mind we probably need to update reference/android-features.md as well given that the Cuttlefish image is going to be supported since 1.30.0

what do you think?
Or we fine-tune that by creating a separate PR?

reference/provided-images.md
```diff
@@ -15,19 +15,25 @@ Anbox Cloud images are regular [Ubuntu cloud images](https://cloud-images.ubuntu
...
-|-----------------------------|----------|-----------------|----------------|---------------|
-| `jammy:aaos15:amd64`        | AAOS     | 15              | 22.04          | 1.26.0 |
-| `jammy:aaos15:arm64`        | AAOS     | 15              | 22.04          | 1.26.0 |
-| `jammy:android15:amd64`     | AOSP     | 15              | 22.04          | 1.26.0 |
-| `jammy:android15:arm64`     | AOSP     | 15              | 22.04          | 1.26.0 |
-| `jammy:aaos14:amd64`        | AAOS     | 14              | 22.04          | 1.24.0 |
-| `jammy:aaos14:arm64`        | AAOS     | 14              | 22.04          | 1.24.0 |
-| `jammy:android14:amd64`     | AOSP     | 14              | 22.04          | 1.24.0 |
-| `jammy:android14:arm64`     | AOSP     | 14              | 22.04          | 1.24.0 |
+| Name                        | Based on | Android Version | Ubuntu Version | Execution model | Available since |
```
adglkh: Execution model

->

Execution environment

tutorial/getting-started-with-virtualized-android.md
ajanon: I went with GB, and removed the space before the unit to be consistent with the rest of the documentation.

We might want to revisit this some time.

reference/feature-support-by-image-type.md
ajanon: I'm not sure if it's worth it to add in the paragraph given that there's a table at the start of the page. I'm removing it from the headers.

adglkh (approved):
LGTM

PR #649, opened 2026-06-18
chore(docs): domain migration

zayatamal:
# Documentation changes

Syncing all domain migration related files (overwrite_links.js)

# Review and preview

In your local repository,

## Reviewers

# JIRA / Launchpad bug


explanation/aaos.md
```diff
@@ -15,7 +15,7 @@ Anbox Cloud offers images based on the [Android Automotive OS (AAOS)](https://so
...
+Until the 1.24.0 release, if you created your own VHAL implementation by following the {ref}`our instructions for customizing the Anbox VHAL <howto-replace-anbox-vhal>` and loaded it as an addon during the Android runtime, vehicle properties and configurations were not visible on the Anbox Cloud dashboard. With the 1.24.0 release, Anbox Cloud has introduced a VHAL adapter that acts as a bridge between the Anbox runtime and the VHAL implementation for data transmission. The Anbox VHAL adapter processes messages sent from the Anbox runtime, performs operations against the VHAL implementation(such as retrieving a list of property configurations), and sends the results back to the Anbox runtime. As a result, vehicle properties and configurations are now visible on the Anbox Cloud dashboard for the third-party VHAL implementations.
```
ajanon: "by following the our"?

I'm guessing 'the' should be removed?

howto/install/validate-deployment.md
```diff
@@ -40,7 +40,7 @@ Do you want to test AMS? [default=yes]
 What is the address of AMS? https://1.2.3.4:8444
 Generating TLS certificate and key for AMS ...
 NOTE: You have to register the certificate generated manually with
- AMS. See https://documentation.ubuntu.com/anbox-cloud/howto/anbox/control-ams-remotely/ for
+ AMS. See howto/anbox/control-ams-remotely.md for
```
ajanon: This shows what the user sees in a terminal, and they will have the entire URL in there.

This should show the entire link, though it probably needs to be updated in the AMS code base first?

_static/js/overwrite_links.js
ajanon, line 1: What is this file about? What's it for?

Neither the commit message nor the code make it very clear to me.

ajanon (changes requested):

howto/install/validate-deployment.md
zayatamal: I understand now, since it redirects to correct domain, ill keep it as the link then. The links will have to be changed to the new domains.

_static/js/overwrite_links.js
zayatamal: this file we added updates links that still point to the old RTD URL and changes them to the new Canonical documentation URL. It ensures that all documentation navigation links continue to use the new domain instead of redirecting users back to the old RTD domain.

explanation/aaos.md
zayatamal: ah yes, i was updating the links and keeping their titles, missed this.

_static/js/overwrite_links.js
ajanon: Okay, but... Where does this happen?
In the doc sources?
In the generated docs? (i.e. _build)
In the hosted rtd website at runtime?

Maybe it's just because I'm unfamiliar with the internals of the starter pack, but it's unclear to me how that works.

zayatamal: I don't know if my answer is clear but:

so basically, the file is in the doc sources but does not modify them (.rst or .md files)

as you can see, the file is added to the Sphinx project (under _static/js/) and is included in the conf.py (in the html_js_files), which tells Sphinx to include a <script> tag for this file in every HTML page.

when RTD builds the documentation, Sphinx copies this added js file to the output with the generated HTML page. RTD then serves this built output.

the actual link rewriting happens at runtime. when a user loads a documentation page, their browser executes the script, which scans the page for links pointing to the old *.readthedocs-hosted.com domain (which our docs is built on on RTD) and replaces them with the new canonical.com/

ajanon: Okay, I think I understand.

Basically, this is a catch-all, juts in case we forgot some links, is that right?

zayatamal: it's not specifically for links we forgot, but more for the links that RTD generates automatically that we have no control over

ajanon (approved):
LGTM

PR #693, opened 2026-07-28
Document the migration guide for addon and app hooks

adglkh:
# Documentation changes

Given the fact that we are dropping application, addon, and registry
management support from the UI starting in version 1.31.0, we will
provide early awareness to end users. This ensures they can begin
migrating ahead of time for a smooth transition before the feature
is completely removed from Anbox Cloud.

# Review and preview

In your local repository,

## Reviewers

# JIRA / Launchpad bug

https://warthogs.atlassian.net/browse/AC-4854

### DIFF
--- a/howto/instance/migrate-from-addon-and-app-hooks.md
+++ b/howto/instance/migrate-from-addon-and-app-hooks.md
@@ -1,0 +1,151 @@
+---
+myst:
+  html_meta:
+    "description": "How to migrate from addon or application hooks (pre-start, post-start, post-stop) to a systemd service built into an Anbox image."
+---
+
+(howto-migrate-from-addon-and-app-hooks)=
+# Migrate from app/addon hooks to system units
+
+```{important}
+Addons and application hooks are deprecated and will be removed from Anbox Cloud. See {ref}`ref-deprecation-notes` for the exact release in which support ends.
+```
+Addon and application hooks (`pre-start`, `post-start`, `post-stop`) are available only for instances with containerized Android (`jammy:*` images) and are not supported for virtualized Android (see {ref}`exp-android-execution-models`).
+
+This guide shows how to replace addon and application hook behavior with a `systemd` unit baked into a custom Anbox image. This approach works for both containerized and virtualized Android and does not depend on the addon subsystem.
+
+## Identify what your hook depends on
+
+Before rewriting your hook as a systemd unit, identify its dependencies. Hook scripts had access to environment variables injected by the Anbox runtime that are not available in systemd. Review your existing hook scripts for the variables below and replace each one with its systemd equivalent.
+
+| Hook variable | Migration |
+| --- | --- |
+| `APP_DIR` | Use `/var/lib/anbox/app` directly. |
+| `ANBOX_DIR` | Use `/var/lib/anbox` directly. |
+| `ANDROID_ROOTFS` | Use `/var/lib/anbox/rootfs` directly. |
+| `BOOT_PACKAGE` | Use the `boot-package` attribute from the application manifest file directly in the script instead of `BOOT_PACKAGE`. |
+| `INSTANCE_TYPE` |  Check whether `/var/lib/anbox/bootstrap.yaml` exists. Its presence indicates a `regular` instance; its absence indicates a `base` instance. |
+| `CONTAINER_TYPE` |This variable is deprecated. Use the same `/var/lib/anbox/bootstrap.yaml `check described for `INSTANCE_TYPE`. |
+| `ANBOX_EXIT_CODE` | Query `anbox.service` directly for the exit code: `systemctl show anbox.service --property=ExecMainStatus`. |
+
+## General workflow
+
+To run a script before or after Android starts, or after Android stops, follow this workflow:
+
+1. Launch an instance from a container image.
+2. Inside that instance, write a `systemd` unit that runs your script at the desired point in the Android life cycle, relative to the `anbox.service` unit, which represents the Anbox runtime.
+3. Enable the unit to automatically activate upon the initial startup or every startup of an instance created from this image.
+4. Publish the modified instance as a new image.
+5. Launch future instances from this custom image instead of relying on an addon or application.
+
+The following sections show concrete examples for each of the three hooks that addons or applications used to provide.
+
+## Launch and access a source instance
+
+Launch an instance to use as the base for your custom image:
+
+    amc launch --name source-inst jammy:android15:amd64
+    amc shell source-inst
+
+All steps below run inside this instance.
+
+### Replace a `pre-start` hook
+
+If you need a script to run on every start before Android starts, create a one-shot unit that starts before `anbox.service` is up and running:
+
+1. Inside the instance, create the unit file, for example `/etc/systemd/system/app-pre-start.service`:
+
+      ```systemd
+      [Unit]
+      Description=Run a script before Android starts
+      Before=anbox.service
+
+      [Service]
+      Type=oneshot
+      ExecStart=/usr/local/bin/pre-start.sh
+      TimeoutStartSec=10min
+
+      [Install]
+      WantedBy=anbox.service
+      ```
+1. Verify the unit file for any syntax or spelling errors:
+
+      sudo systemd-analyze verify /etc/systemd/system/app-pre-start.service
+
+1. Add your script (`/usr/local/bin/pre-start.sh` in this example), make it executable, then enable the unit.
+
+      sudo chmod +x /usr/local/bin/pre-start.sh
+      sudo systemctl daemon-reload
+      sudo systemctl enable app-pre-start.service
+
+```{note}
+AMS waits for an instance to become fully up and running with a maximum timeout of 15 minutes. Consequently, all scripts should be as lightweight as possible and avoid long-running operations. We recommend setting an appropriate `TimeoutStartSec` in the unit file to prevent instance launches from failing due to long-running operations in the pre-start unit.
+```
+
+### Replace a `post-start` hook
+
+If you need a script to run every time after Android starts, create a one-shot unit that starts after anbox.service is up and running.
+
+1. Inside the instance, create the unit file, for example `/etc/systemd/system/app-post-start.service`:
+
+      ```systemd
+      [Unit]
+      Description=Run a script after Android has started
+      After=anbox.service
+      Requires=anbox.service
+
+      [Service]
+      Type=oneshot
+      ExecStart=/usr/local/bin/post-start.sh
+
+      [Install]
+      WantedBy=multi-user.target
+      ```
+
+1. Add your script (`/usr/local/bin/post-start.sh` in this example), make it executable, and enable it the same way as in the previous section.
+
+### Replace a `post-stop` hook
+
+If you need a script to run after Android was stopped, bind a unit to the life cycle of `anbox.service` and run your cleanup or data backup logic in `ExecStop`, so it executes whenever `anbox.service` stops:
+
+1. Inside the instance, create the unit file, for example `/etc/systemd/system/app-post-stop.service`:
+
+      ```systemd
+      [Unit]
+      Description=Run a script after Android has stopped
+      Before=anbox.service
+      BindsTo=anbox.service
+
+      [Service]
+      Type=oneshot
+      ExecStop=/usr/local/bin/post-stop.sh
+      RemainAfterExit=yes
+
+      [Install]
+      WantedBy=anbox.service
+      ```
+
+1. Add your script (`/usr/local/bin/post-stop.sh` in this example), make it executable, and enable it the same way as in the previous section.
+
+## Publish and use custom image
+
+Once your units and scripts are in place inside the instance, publish it as a new image(see {ref}`howto-publish-instance-as-image`) and use it for future instances:
+
+    amc publish source-inst --name custom-image --force
+    amc launch --name new-instance custom-image
+
+Verify that your system units ran as expected by checking the instance logs(see {ref}`howto-view-instance-logs`),
+
+    amc logs new-instance
+
+Or check the logs of a specific system unit,
+
+    amc exec new-instance -- journalctl -u app-pre-start.service
+
+## Related topics
+
+- {ref}`ref-hooks`
+- {ref}`ref-deprecation-notes`
+- {ref}`howto-publish-instance-as-image`
+- {ref}`howto-create-instance`
+- {ref}`howto-view-instance-logs`
--- a/reference/deprecation-notices.md
+++ b/reference/deprecation-notices.md
@@ -86,3 +86,5 @@ This transition provides a two-year migration window before legacy charms become
 Support for [applications](/howto/application/index.md), [addons](/howto/addons/index.md), and Anbox Application Registry [AAR](/howto/aar/index.md) is deprecated as of version 1.31.0.
 - In 1.31.0: Creating applications and registry management via the Anbox Cloud dashboard is disabled. However, applications, addons, and registry management remains functional through the CLI for transition purposes.
 - In 1.32.0: All support for applications, addons and registry will be officially removed. Users will no longer be able to create, manage, or utilize these features via the CLI or dashboard.
+
+See {ref}`howto-migrate-from-addon-and-app-hooks` for how to replicate addon hook behavior using a custom image instead.
--- a/reference/deprecation-notices.md
+++ b/reference/deprecation-notices.md
@@ -81,10 +81,10 @@ Starting from 1.29.0, all new Anbox Cloud deployments must use modernized operat
 This transition provides a two-year migration window before legacy charms become completely unsupported. Anbox Cloud 1.34.2 will be the last release that supports deployments using legacy charms. Starting from Anbox Cloud 1.35.0, Anbox Cloud deployments will exclusively support modernized charms.
 
 ## Applications, Addons and Registry in AMS
-*Deprecated in 1.31.0* ; *Unsupported in 1.32.0*
+*Deprecated in 1.32.0* ; *Unsupported in 1.34.0*
 
 Support for [applications](/howto/application/index.md), [addons](/howto/addons/index.md), and Anbox Application Registry [AAR](/howto/aar/index.md) is deprecated as of version 1.31.0.
-- In 1.31.0: Creating applications and registry management via the Anbox Cloud dashboard is disabled. However, applications, addons, and registry management remains functional through the CLI for transition purposes.
-- In 1.32.0: All support for applications, addons and registry will be officially removed. Users will no longer be able to create, manage, or utilize these features via the CLI or dashboard.
+- In 1.32.0: Creating applications and registry management via the Anbox Cloud dashboard is disabled. However, applications, addons, and registry management remains functional through the CLI for transition purposes.
+- In 1.34.0: All support for applications, addons and registry will be officially removed. Users will no longer be able to create, manage, or utilize these features via the CLI or dashboard.
 
 See {ref}`howto-migrate-from-addon-and-app-hooks` for how to replicate addon hook behavior using a custom image instead.
--- a/.custom_wordlist.txt
+++ b/.custom_wordlist.txt
@@ -407,6 +407,8 @@ YUV
 ZFS
 zpool
 zsh
+chmod
+systemctl
 
 # To be removed after AMS sync
 authrorization
--- a/explanation/addons.md
+++ b/explanation/addons.md
@@ -7,6 +7,10 @@ myst:
 (exp-addons)=
 # Addons
 
+```{caution}
+Addons including hooks, are deprecated and are being removed from Anbox Cloud. See {ref}`ref-deprecation-notes` for the removal schedule and {ref}`howto-migrate-from-addon-and-app-hooks` for how to replicate hook behavior without addons.
+```
+
 Addons provide a way to extend and customize images in Anbox Cloud. Once you have created addons, you can create hooks for them that are triggered based on events in the life cycle of an instance. You can create addons independently and later attach it to individual applications.
 
 ## Best practices while using addons
--- a/explanation/applications.md
+++ b/explanation/applications.md
@@ -7,6 +7,10 @@ myst:
 (exp-applications)=
 # Applications
 
+```{caution}
+Applications including hooks, are deprecated and are being removed from Anbox Cloud. See {ref}`ref-deprecation-notes` for the removal schedule and {ref}`howto-migrate-from-addon-and-app-hooks` for how to replicate hook behavior without addons.
+```
+
 Applications are one of the main objects managed by Anbox Management Service (AMS). A single application encapsulates one Android APK ([Android Package Kit](https://en.wikipedia.org/wiki/Android_application_package)) and manages it within the cluster. It takes care of installing the supplied APK and making it available to users. AMS also manages updates to existing applications, which includes allowing the operator to test new uploaded versions before making them available to any users.
 
 ## Application requirements
--- a/howto/instance/index.md
+++ b/howto/instance/index.md
@@ -66,4 +66,5 @@ delete-instance
 backup-restore-application-data
 configure-geographic-location
 expose-services
+migrate-from-addon-and-app-hooks
 ```
--- a/reference/hooks.md
+++ b/reference/hooks.md
@@ -7,6 +7,10 @@ myst:
 (ref-hooks)=
 # Hooks
 
+```{caution}
+Addons and apps, including hooks, are deprecated and are being removed from Anbox Cloud. See {ref}`ref-deprecation-notes` for the removal schedule and {ref}`howto-migrate-from-addon-and-app-hooks` for how to replicate hook behavior without addons.
+```
+
 Hooks are scripts that automatically trigger actions based on an event performed in the life cycle of an instance. A hook can be any executable file that is placed in the `hooks` directory of an addon or an application folder.
 
 The hook name **must** be one of the following:


reference/deprecation-notices.md
```diff
@@ -84,5 +84,7 @@ This transition provides a two-year migration window before legacy charms become
 *Deprecated in 1.31.0* ; *Unsupported in 1.32.0*
 
 Support for [applications](/howto/application/index.md), [addons](/howto/addons/index.md), and Anbox Application Registry [AAR](/howto/aar/index.md) is deprecated as of version 1.31.0.
-- In 1.31.0: Creating applications and registry management via the Anbox Cloud dashboard is disabled. However, applications, addons, and registry management remains functional through the CLI for transition purposes.
-- In 1.32.0: All support for applications, addons and registry will be officially removed. Users will no longer be able to create, manage, or utilize these features via the CLI or dashboard.
+- In 1.32.0: Creating applications and registry management via the Anbox Cloud dashboard is disabled. However, applications, addons, and registry management remains functional through the CLI for transition purposes.
```
ajanon, line 87: I wasn't aware we were changing the timeline?

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
...
+They are not supported for instances with virtualized Android, see {ref}`exp-android-execution-models`.
+
+Addons and applications, including their hooks, are deprecated and are being removed from Anbox Cloud.
+See {ref}`ref-deprecation-notes` for the exact release in which support ends.
+
+This guide describes how to reproduce the behavior of addon hooks without relying on addons, by baking the equivalent logic directly into an Anbox image using a `systemd` unit.
+
+## Why use this approach
+
+An hook only ever runs the same script at the same point in the instance life cycle, for every instance launched from an Anbox image.
```
ajanon, line 20: Is it 'a hook' or 'an hook'?

I think there's a rule, but it's kinda unclear to me... Maybe we wait until Amal is back to have confirmation?

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
...
+
+3. Add your script (`/usr/local/bin/pre-start.sh` in this example), make it executable, then enable the unit:
+
+      sudo chmod +x /usr/local/bin/pre-start.sh
+      sudo systemctl daemon-reload
+      sudo systemctl enable app-pre-start.service
+
+```{note}
+AMS waits for an instance to become fully up and running with a maximum timeout of **15 minutes**.
+Consequently, all scripts should be as lightweight as possible and avoid long-running operations. We recommend setting an appropriate `TimeoutStartSec` in the unit file to prevent instance launches from failing due to long-running operations in the pre-start unit.
```
ajanon, line 69:
````suggestion
```{note}
AMS waits for an instance to become fully up and running with a maximum timeout of **15 minutes**.
Consequently, all scripts should be as lightweight as possible and avoid long-running operations. We recommend setting an appropriate `TimeoutStartSec` in the unit file to prevent instance launches from failing due to long-running operations in the pre-start unit.
```
````

File: howto/instance/migrate-from-addon-and-app-hooks.md [hunk 4]
```diff
@@ -0,0 +1,131 @@
...
+
+      amc launch --name source-inst jammy:android15:amd64
+      amc shell source-inst
+
+2. Inside the instance, create the unit file, for example `/etc/systemd/system/app-pre-start.service`:
+
+      ```systemd
+      [Unit]
+      Description=Run a script before Android starts
+      Before=anbox.service
```
ajanon, line 50: We probably needs Requires?
```suggestion
      Before=anbox.service
      Requires=anbox.service
```

That will protect against running in unsupported images.

ajanon (changes requested):

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: It should be "a hook" and not "an hook". The simple rule is it starts with a consonant **sound** (/h/) so "a"

reference/deprecation-notices.md
```diff
@@ -84,5 +84,7 @@ This transition provides a two-year migration window before legacy charms become
 *Deprecated in 1.31.0* ; *Unsupported in 1.32.0*
```
zayatamal: The body below updates the information to say:
- 1.32.0 (dashboard disabled) but functional through CLI
- 1.34.0 (full removal)
but the existing header here still reads: *Unsupported in 1.32.0*
This must be updated to *Unsupported in 1.34.0* to match the new timeline in the body.

Another question: when and where did we say it is going to be in 1.34.0

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
...
+      sudo systemctl daemon-reload
+      sudo systemctl enable app-pre-start.service
+
+```{note}
+AMS waits for an instance to become fully up and running with a maximum timeout of **15 minutes**.
+Consequently, all scripts should be as lightweight as possible and avoid long-running operations. We recommend setting an appropriate `TimeoutStartSec` in the unit file to prevent instance launches from failing due to long-running operations in the pre-start unit.
+
+## Replace a `post-start` hook
+
+If you need a script to run every time after Android starts, create a one-shot unit that starts after anbox.service is up and running."
```
zayatamal: minor but there is a trailing " that doesn't open any quote. Should be dropped

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
...
+
+## General workflow
+
+To run a script before or after Android starts, or after Android stops, follow this workflow:
+
+1. Launch an instance from a container image.
+2. Inside that instance, write a `systemd` unit that runs your script at the desired point in the Android life cycle, relative to the `anbox.service` unit, which represents the Anbox runtime.
+3. Enable the unit to automatically activate upon the initial startup or every startup of an instance created from this image.
+4. Publish the modified instance as a new image.
+5. Launch future instances from this custom image instead of relying on an addon or application
```
zayatamal: minor: missing a period

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: I think @ajanon is right here!

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal, line 1: I have a question(s) about the placement of this how to guide:
It's currently under manage instances but wouldn't make more sense if (I have two options):
- either in the **manage addons** **BUT**  this guide teaches users to leave addons behind entirely, which means when addons are fully removed (1.34.0), that entire section will be pruned or hidden: taking the migration guide with it. Users who haven't migrated yet would lose access to the instructions. though we can have a deprecation banner throughout the full transition window.

- or under **manage images** section and my reason for this based on reading the guide is the process that is happening ad well as the guide's output being a custom image:
1. Launch an instance\
2. Add systemd units + scripts
3. Publish as a custom image (amc publish)
4. Use that image going forward 

The other reason i brought this up because currently: the manage instances guides are about managing instance lifecycle (create, stop, start, delete), not about what runs inside them.

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
+---
+myst:
+  html_meta:
+    "description": "How to migrate from addon or application hooks (pre-start, post-start, post-stop) to a systemd service built into an Anbox image."
+---
+
+(howto-migrate-from-addon-and-app-hooks)=
+# Migrate from addon hooks
```
zayatamal: The introductory paragraph below is too long for a how-to guide and is slightly verbose, maybe something along the lines of:

Addon hooks (`pre-start`, `post-start` and `post-stop`, see {ref}`ref-hooks`) are deprecated and will be removed from Anbox Cloud in version 1.34.0 (see {ref}`ref-deprecation-notes`). They are only available for instances with containerized Android and are not supported for virtualized Android (see {ref}`exp-android-execution-models`).

This guide shows how to replace addon hooks with `systemd` units baked into a custom image. A `systemd` unit runs at the same lifecycle point as the hook it replaces, works for both containerized and virtualized Android and is the standard approach for customizing instance startup and shutdown going forward.

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
...
+
+Addon hooks (`pre-start`, `post-start` and `post-stop`, see {ref}`ref-hooks`) are only available for instances with containerized Android (`jammy:*` images).
+They are not supported for instances with virtualized Android, see {ref}`exp-android-execution-models`.
+
+Addons and applications, including their hooks, are deprecated and are being removed from Anbox Cloud.
+See {ref}`ref-deprecation-notes` for the exact release in which support ends.
+
+This guide describes how to reproduce the behavior of addon hooks without relying on addons, by baking the equivalent logic directly into an Anbox image using a `systemd` unit.
+
+## Why use this approach
```
zayatamal: this section would be removed as it's added above in the intro, my justification:
Why use this approach functions as an explanation and it answers a WHY so it breaks the how-to pattern. the reason that I find no exception for it to be here is when the reader faces a genuine choice between multiple paths and needs guidance to pick one. Diátaxis acknowledges that brief orientation is acceptable when a decision point exists.

But here, there is no choice. Addons are being removed. There's exactly one path forward: systemd units in a custom image. The reader doesn't need to be convinced, they need to be shown how.

and we already incorporated it in the intro above.

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,131 @@
...
+
+To run a script before or after Android starts, or after Android stops, follow this workflow:
+
+1. Launch an instance from a container image.
+2. Inside that instance, write a `systemd` unit that runs your script at the desired point in the Android life cycle, relative to the `anbox.service` unit, which represents the Anbox runtime.
+3. Enable the unit to automatically activate upon the initial startup or every startup of an instance created from this image.
+4. Publish the modified instance as a new image.
+5. Launch future instances from this custom image instead of relying on an addon or application
+
+The following sections show concrete examples for each of the three hooks that addons or applications used to provide.
```
zayatamal, line 41: since the first step in the general overview is:  Launch an instance from a container image. and we have it added only for ## Replace a `pre-start` hook, here's what we can do: 
we can have a section for all 3 before the three examples:
## Launch and access a source instance

Launch an instance to use as the base for your custom image:

    amc launch --name source-inst jammy:android15:amd64
    amc shell source-inst

All steps below run inside this instance.


based on this we can remove the 1st step in the ## Replace a `pre-start` hook

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: what about this for a heading: # Replace addon hooks with systemd units ?

reason: states both origin and destination

reference/deprecation-notices.md
adglkh: We were.
See
https://docs.google.com/document/d/1DUSsSmAtEJwY1DRCPK5mLtTra0gg6pOxRyZAMp6KPj8/edit?tab=t.0#heading=h.r8rjo665pzy0

reference/deprecation-notices.md
adglkh: > This must be updated to Unsupported in 1.34.0 to match the new timeline in the body.

A: +1, thanks for pointing this out.

> Another question: when and where did we say it is going to be in 1.34.0

A: People can check the exact timeline on the deprecation notices page. Apart from that, we can also mention this in the 1.32.0 release notes(so in the 1.32.0 release timeline)

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: +1, fixed that.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Fixed that.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Thanks, I fixed it.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Make a lots sense to me.
Thanks

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: My fault.
Fixed that.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: +1, thanks for the suggestion. I polished the intro a bit with more context.
Please take a look.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: I'd prefer to keep it as is. 'Replace addon hooks with systemd units' sounds a bit like a two-way choice, whereas this guide is specifically meant to facilitate a mandatory migration path to systemd units.

what do you think?

Also, regarding the heading, probably the following sounds better 
# Migrate from app/addon hooks to system units

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Addon/app hooks and the new standard approach (using systemd units to run a hook at specific points in the instance lifecycle) both act on the instance rather than the image or managing addons.
So I lean toward to keep this page under the `manage instances`.

However, this brings up another consideration: even after 1.34.0 is released, we probably need to keep this guide active for at least one more release in case people miss the deprecation notice. Meaning we'll still need this page post-1.34.0. In 1.35.0, we can rework/evolve this page, perhaps renaming it to something like `Running a script at specific instance lifecycle points`.

what do you think? @zayatamal @ajanon

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: Yes we will go with your heading :+1:

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: hmm, i don't fully understand the point `That will protect against running in unsupported images`. This is anbox image which comes with anbox service. We're in a specific boot/test sequence. `Before=anbox.service` is already enough to guarantee the ordering. Do we really need a hard dependency here?

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,129 @@
+---
+myst:
+  html_meta:
+    "description": "How to migrate from addon or application hooks (pre-start, post-start, post-stop) to a systemd service built into an Anbox image."
+---
+
+(howto-migrate-from-addon-and-app-hooks)=
+# Migrate from app/addon hooks to system units
```
zayatamal, line 8: the intro still feels a bit long and needs more tightening, and mainly for redundancy I would suggest having an important admonition:
i want it to tell me the important info first which is that it's being deprecated hence the existence of this guide and as a user following it<< this is going away, check the timeline and act:

```{important}
Addons and application hooks are deprecated and will be removed from Anbox Cloud. See {ref}`ref-deprecation-notes` for the exact release in which support ends.
```
Addon and application hooks (`pre-start`, `post-start`, `post-stop`) are available only for instances with containerized Android (jammy:* images) and are not supported for virtualized Android (see {ref}exp-android-execution-models).

This guide shows how to replace addon and application hook behavior with a `systemd` unit baked into a custom Anbox image. This approach works for both containerized and virtualized Android and does not depend on the addon subsystem.
^^^^
the reason i did this is because
**this** " A `systemd` unit inside the image achieves the same outcome without depending on the addon subsystem. This makes it the standard way to customize instance startup and shutdown going forward for both containerized and virtualized Android." **and this** "to reproduce the behavior of addon hooks without relying on addons, by baking the equivalent logic directly into an Anbox image using a `systemd` unit." are in a way saying the same thing

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,129 @@
...
+Addon hooks (`pre-start`, `post-start` and `post-stop`, see {ref}`ref-hooks`) are only available for instances with containerized Android (`jammy:*` images).
+They are not supported for instances with virtualized Android, see {ref}`exp-android-execution-models`.
+
+Addons and applications, including their hooks, are deprecated and are being removed from Anbox Cloud. A `systemd` unit inside the image achieves the same outcome without depending on the addon subsystem. This makes it the standard way to customize instance startup and shutdown going forward for both containerized and virtualized Android.
+See {ref}`ref-deprecation-notes` for the exact release in which support ends.
+
+This guide describes how to reproduce the behavior of addon hooks without relying on addons, by baking the equivalent logic directly into an Anbox image using a `systemd` unit.
+
+
+## General workflow
```
zayatamal, line 31: we can ommit the general workflow as the section headings already communicate the sequence clearly, and the introduction explains the goal. The overview just restates what the reader will encounter. so in a way it's adding a chunck of the same information/process below

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,129 @@
...
+
+      [Install]
+      WantedBy=anbox.service
+      ```
+
+1. Add your script (`/usr/local/bin/post-stop.sh` in this example), make it executable, and enable it the same way as in the previous section.
+
+## Publish and use custom image
+
+Once your units and scripts are in place inside the instance, publish it as a new image and use it for future instances:
```
zayatamal: we can add a cross reference here for publishing as a new image: See {ref}howto-publish-instance-as-image for detailed steps.

and maybe have it as: 
Once your units and scripts are in place inside the instance, publish it as a new image and use it for future instances. See {ref}howto-publish-instance-as-image for detailed steps.
amc publish source-inst --name custom-image --force

Launch a new instance from the custom image:
  amc launch --name new-instance custom-image

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,129 @@
...
+Launch an instance to use as the base for your custom image:
+
+    amc launch --name source-inst jammy:android15:amd64
+    amc shell source-inst
+
+All steps below run inside this instance.
+
+### Replace a `pre-start` hook
+
+If you need a script to run on every start before Android starts: create a one-shot unit that starts before `anbox.service` is up and running:
```
zayatamal: If you need a script to run on every start before Android starts:
replace ":" with ","

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,129 @@
...
+      [Service]
+      Type=oneshot
+      ExecStart=/usr/local/bin/pre-start.sh
+      TimeoutStartSec=10min
+
+      [Install]
+      WantedBy=anbox.service
+      ```
+
+1. Add your script (`/usr/local/bin/pre-start.sh` in this example), make it executable, then enable the unit:
```
zayatamal: make sure the number list renders correctly when you do `make run`

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Okay, convinced. 
Let's go with your suggestion here.

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,129 @@
...
+      ExecStop=/usr/local/bin/post-stop.sh
+      RemainAfterExit=yes
+
+      [Install]
+      WantedBy=anbox.service
+      ```
+
+1. Add your script (`/usr/local/bin/post-stop.sh` in this example), make it executable, and enable it the same way as in the previous section.
+
+## Publish and use custom image
```
zayatamal, line 130: is there a way to verify before publishing to confirm that the unit is enabled and runs correctly?

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: +1, make sense to me.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Fixed it.

zayatamal (commented):
@adglkh I have a question by referring to the addons guides:
would `INSTANCE_TYPE` , `ADDON_DIR` be affected or should we have a process or mention of them since these are not  not available in a systemd context

if they're relevant, no longer relevant, because users migrating from hooks will have scripts that reference these variables.

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: Okay I see your point regarding manage instance. and regarding the second point, yes I also agree about keeping it active for a while  in case people miss the deprecation notice. 

there's the possibilty of users with deprecated addons looking in the addons section first so we can have a cross-reference in the addons index page or any page of relevance.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: 1. If the unit isn't enabled correctly, the user will encounter corresponding errors when running `sudo systemctl enable <unit_service>`. Also, we can't prevent users from adding other directives in their unit files using standard systemd syntax， they'll need to check for any related errors themselves.

Probably we could polish the section
```
+1. Add your script (`/usr/local/bin/pre-start.sh` in this example), make it executable, then enable the unit:
+
+      sudo chmod +x /usr/local/bin/pre-start.sh
+      sudo systemctl daemon-reload
+      sudo systemctl enable app-pre-start.service
```

a bit for the clarity.

2. To verify if the system unit runs correctly, people could do a
`
journalctl -u app-pre-start.service
`
check the system journal for any syntax or configuration errors:

adglkh:
> would `INSTANCE_TYPE` , `ADDON_DIR` be affected or should we have a process or mention of them since these are not not available in a systemd context
> 
> if they're relevant, no longer relevant, because users migrating from hooks will have scripts that reference these variables.

A: That's a good point. Given that we've already listed all available environment variables natively exported to the hook runtime in reference/hooks.md, complete with their respective values so people can easily map and use them in their hooks.

The only question is whether we should direct users to reference/hooks.md or just create a table directly in the migration guide. I vote for the latter (even if it looks a bit redundant) for the following reasons:
1. The reference/hooks.md will be removed anyway.
2. A couple of variables (e.g., BOOT_PACKAGE / CONTAINER_TYPE) are deprecated, so we can take this opportunity to clean things up.

what do you think? @zayatamal

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: Okay this polish would work, just a minor "." instead of : at the end

zayatamal:
> > would `INSTANCE_TYPE` , `ADDON_DIR` be affected or should we have a process or mention of them since these are not not available in a systemd context
> > if they're relevant, no longer relevant, because users migrating from hooks will have scripts that reference these variables.
> 
> A: That's a good point. Given that we've already listed all available environment variables natively exported to the hook runtime in reference/hooks.md, complete with their respective values so people can easily map and use them in their hooks.
> 
> The only question is whether we should direct users to reference/hooks.md or just create a table directly in the migration guide. I vote for the latter (even if it looks a bit redundant) for the following reasons:
> 
> 1. The reference/hooks.md will be removed anyway.
> 2. A couple of variables (e.g., BOOT_PACKAGE / CONTAINER_TYPE) are deprecated, so we can take this opportunity to clean things up.
> 
> what do you think? @zayatamal

having the table sounds a good plan given that the hooks reference page will be removed. but we need to pay attention to accuracy and in case we miss anything or if the table alone isn't clear enough on how to deal with the cases.  which means we need to frame the table as "Hook environment variables and their migration path"

What about instead of just adding  the table have something that would give the users some context before the launch and acsess step:

Before rewriting your hook as a systemd unit, identify what your hook depended on. Hook scripts had access to environment variables injected by AMS that do not exist in a systemd unit.
a small section or line for `INSTANCE_TYPE` 
a small section or line for `ADDON_DIR`
and other paths
how does that sound @adglkh ?

howto/instance/migrate-from-addon-and-app-hooks.md
ajanon: If you're against it, we can keep it without 'Requires', I'm fine with it.

adglkh: Let's keep it as is then.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: > `...  we can have a cross-reference in the addons index page or any page of relevance.`

A: +1, let me add that.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Fixed that.

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: hmm, I see your point. However, I kept it as a quick high-level summary so readers can get an overview of the steps before diving into the details. What do you think?

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Writing `1.` for every item in a Markdown ordered list will correctly render as a consecutive, ascending sequence. I don't see anything wrong after it's rendered.
 
See https://jmmv.dev/2022/07/markdown-lists.html

adglkh:
> > > would `INSTANCE_TYPE` , `ADDON_DIR` be affected or should we have a process or mention of them since these are not not available in a systemd context
> > > if they're relevant, no longer relevant, because users migrating from hooks will have scripts that reference these variables.
> > 
> > 
> > A: That's a good point. Given that we've already listed all available environment variables natively exported to the hook runtime in reference/hooks.md, complete with their respective values so people can easily map and use them in their hooks.
> > The only question is whether we should direct users to reference/hooks.md or just create a table directly in the migration guide. I vote for the latter (even if it looks a bit redundant) for the following reasons:
> > 
> > 1. The reference/hooks.md will be removed anyway.
> > 2. A couple of variables (e.g., BOOT_PACKAGE / CONTAINER_TYPE) are deprecated, so we can take this opportunity to clean things up.
> > 
> > what do you think? @zayatamal
> 
> having the table sounds a good plan given that the hooks reference page will be removed. but we need to pay attention to accuracy and in case we miss anything or if the table alone isn't clear enough on how to deal with the cases. which means we need to frame the table as "Hook environment variables and their migration path"
> 
> What about instead of just adding the table have something that would give the users some context before the launch and acsess step:
> 
> Before rewriting your hook as a systemd unit, identify what your hook depended on. Hook scripts had access to environment variables injected by AMS that do not exist in a systemd unit. a small section or line for `INSTANCE_TYPE` a small section or line for `ADDON_DIR` and other paths how does that sound @adglkh ?

A: Regarding `What about instead of just adding the table have something that would give the users some context before the launch and acsess step`

Those env vars are not injected by AMS, but rather by an internal tool (acc) running inside the classic container image, so those environment variables would have been phased out anyway. I'd suggest a table for each path, but with a dedicated column mapping out the old env variables. Once 1.34.0 is out, we could polish this table by simply stripping out that column.

What do you think?

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,167 @@
...
+    amc publish source-inst --name custom-image --force
+    amc launch --name new-instance custom-image
+
+Verify that your system units ran as expected by checking the instance logs(see {ref}`howto-view-instance-logs`),
+
+    amc logs new-instance
+
+Or check the logs of a specific system unit,
+
+    amc exec new -- journalctl -u app-pre-start.service
```
zayatamal: @adglkh shouldnt this be: amc exec new-instance -- journalctl -u app-pre-start.service

howto/instance/migrate-from-addon-and-app-hooks.md
zayatamal: we can keep it for now

howto/instance/migrate-from-addon-and-app-hooks.md
```diff
@@ -0,0 +1,167 @@
...
+# Migrate from app/addon hooks to system units
+
+```{important}
+Addons and application hooks are deprecated and will be removed from Anbox Cloud. See {ref}`ref-deprecation-notes` for the exact release in which support ends.
+```
+Addon and application hooks (`pre-start`, `post-start`, `post-stop`) are available only for instances with containerized Android (`jammy:*` images) and are not supported for virtualized Android (see {ref}`exp-android-execution-models`).
+
+This guide shows how to replace addon and application hook behavior with a `systemd` unit baked into a custom Anbox image. This approach works for both containerized and virtualized Android and does not depend on the addon subsystem.
+
+## Identify what your hook depends on
```
zayatamal, line 17: @adglkh after reading this, I think your instinct about having a table here was correct, that makes the table a decision/mapping step, maybe we can have something like the following (i mapped out the sentences to a table):
## Migrate existing hook logic

Before rewriting your hook as a systemd unit, identify its dependencies. Hook scripts had access to environment variables injected by the Anbox runtime that are not available in systemd. Review your existing hook scripts for the variables below and replace each one with its systemd equivalent.

| Hook variable | Migration |
| --- | --- |
| `APP_DIR` | Use `/var/lib/anbox/app` directly. |
| `ANBOX_DIR` | Use `/var/lib/anbox` directly. |
| `ANDROID_ROOTFS` | Use `/var/lib/anbox/rootfs` directly. |
| `BOOT_PACKAGE` | Use the `boot-package` attribute directly in the script instead of `BOOT_PACKAGE`.|
| `INSTANCE_TYPE` |  Check whether `/var/lib/anbox/bootstrap.yaml` exists. Its presence indicates a `regular` instance; its absence indicates a `base` instance.|
| `CONTAINER_TYPE` |This variable is deprecated. Use the same `/var/lib/anbox/bootstrap.yaml `check described for `INSTANCE_TYPE`.|
| `ANBOX_EXIT_CODE` | Query `anbox.service` directly for the exit code: `systemctl show anbox.service --property=ExecMainStatus`. |


other than this part, I have no more comments

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: Ah, yes. Thanks for pointing this out

howto/instance/migrate-from-addon-and-app-hooks.md
adglkh: +1, a table would be neat in such a case.

